PR #2, opened 2017-03-09
Create workflow prototype

mohit:
What do we have here:

This PR implements a basic API to 
1. `Create` and `Update` `Workflows`
2. `Start` and `GetStatus` for `Jobs` which are instantiations of `Workflows`

does not contain generated files. You can find them at https://github.com/Clever/workflow-manager/pull/3

The major packages involved are:

* `main`: contains the api `handler` 

* `resources`: structs for the major concepts used by the `workflow-manager` namely

  * WorkflowDefinition
  * State
  * Job
  * Task

Also contains `KitchenSink` which is a workflow resource that can be used by other packages in tests. 

* `executor`: contains JobManager and a wrapper over `aws-go-sdk`'s Batch API
  * Yes this needs better naming but this is just easier to work with now, and should be easy to refactor.
  * Suggestions for better naming and organization welcome!
  * `JobManager` currently is currently a place where Workflows are converted to tasks and vice versa.
  *  `JobManager` also converts between the resources to details that can be consumed by the `BatchClient` 

* `store` defines and interface for the store and contains a very simple in-memory implementation


resources/workflows.go
```diff
@@ -0,0 +1,182 @@
...
+	IsEnd() bool
+}
+
+// Workflow defines an interface for defining a flow of States
+type Workflow interface {
+	Name() string
+	Version() int
+	StartAt() State
+	States() map[string]State
+	//NextState(stateName string) State
```
burnsed: delete the commented code?

resources/workflows.go
```diff
@@ -0,0 +1,182 @@
...
+
+	"github.com/Clever/workflow-manager/toposort"
+)
+
+// State refers to the different states in a Workflow
+type State interface {
+	Name() string
+	Type() string
+	Next() string
+	Dependencies() []string
```
burnsed: Is it worth separating out the fields (like `Dependencies`) that are populated internally vs by users?

burnsed: or maybe marking them differently?

resources/workflows.go
```diff
@@ -0,0 +1,182 @@
...
+	AddDependency(State)
+	Resource() string
+	IsEnd() bool
```
burnsed: NIT: `IsLastState` might be more clear

resources/workflows.go
```diff
@@ -0,0 +1,182 @@
...
+}
+
+// Workflow defines an interface for defining a flow of States
+type Workflow interface {
+	Name() string
+	Version() int
+	StartAt() State
```
burnsed: Does the presence of `StartAt` imply that a user could start a job midway through the workflow?

resources/tasks.go
```diff
@@ -0,0 +1,66 @@
...
+	TaskStatusFailed      TaskStatus = "FAILED"              // failed
+	TaskStatusAborted     TaskStatus = "ABORTED_DEPS_FAILED" // aborted due to a dependency failure
+	TaskStatusUserAborted TaskStatus = "ABORTED_BY_USER"     // aborted due to user action
+)
+
+type TaskDetail struct {
+	CreatedAt    time.Time
+	StartedAt    time.Time
+	StoppedAt    time.Time
+	Container    string
```
burnsed: Does this container the full image and tag as a single string?

launch/workflow-manager.yml
```diff
@@ -1,10 +1,13 @@
 env:
-- LIGHTSTEP_ACCESS_TOKEN
+  # AWS Batch settings and credentials
+  - AWS_REGION
+  - BATCH_QUEUE
+  - AWS_ACCESS_KEY_ID
+  - AWS_SECRET_ACCESS_KEY
+#- LIGHTSTEP_ACCESS_TOKEN
```
burnsed: remove the commented out env variable

launch/workflow-manager.yml
```diff
@@ -1,10 +1,13 @@
 env:
-- LIGHTSTEP_ACCESS_TOKEN
+  # AWS Batch settings and credentials
+  - AWS_REGION
```
burnsed: what about `AWS_PROFILE` ?

burnsed: oh, looks like it is optional. Maybe it would be good to include as a commented out item with a description

launch/workflow-manager.yml
```diff
@@ -1,10 +1,13 @@
...
 resources:
-  cpu: 0.1
-  max_mem: 0.1
-shepherds:
-- "mohit.gupta@clever.com"
+  cpu: 0.01
+  max_mem: 0.02
 expose:
 - name: default
   port: 80
```
burnsed: Where does `port` get set in the app?

handler.go
```diff
@@ -18,9 +25,195 @@ func (wm WorkflowManager) HealthCheck(ctx context.Context) error {
 }
 
 // NewWorkflow creates a new workflow
-func (wm WorkflowManager) NewWorkflow(ctx context.Context) (*models.NewWorkflowResponse, error) {
+func (wm WorkflowManager) NewWorkflow(ctx context.Context, workflowReq *models.NewWorkflowRequest) (*models.Workflow, error) {
+
+	if len(workflowReq.States) == 0 || workflowReq.Name == "" {
+		return &models.Workflow{}, fmt.Errorf("Must define at least one state")
```
burnsed: NIT: error message refers to states, but may be due to missing `name`

handler.go
```diff
@@ -18,9 +25,195 @@ func (wm WorkflowManager) HealthCheck(ctx context.Context) error {
...
+
+	return apiWorkflowFromStore(workflow), nil
+}
+
+func (wm WorkflowManager) UpdateWorkflow(ctx context.Context, input *models.UpdateWorkflowInput) (*models.Workflow, error) {
+	if input.NewWorkflowRequest == nil || input.NewWorkflowRequest.Name != input.Name {
+		return &models.Workflow{}, fmt.Errorf("Name in path must match Workflow object")
+	}
+
+	workflowReq := input.NewWorkflowRequest
```
burnsed: move this line to the top and use it in the first set of checks

handler.go
```diff
@@ -18,9 +25,195 @@ func (wm WorkflowManager) HealthCheck(ctx context.Context) error {
...
+	if len(workflowReq.States) == 0 || workflowReq.Name == "" {
```
burnsed: `name` has already been validated

handler.go
```diff
@@ -18,9 +25,195 @@ func (wm WorkflowManager) HealthCheck(ctx context.Context) error {
...
+func (wm WorkflowManager) StartJobForWorkflow(ctx context.Context, input *models.JobInput) (*models.Job, error) {
+	// TODO: also support input.Workflow.Revision
+	workflow, err := wm.store.LatestWorkflow(input.Workflow.Name)
+	if err != nil {
+		return &models.Job{}, err
+	}
+
+	var data []string
+	if input.Data != nil {
+		data = jsonToArgs(input.Data)
```
burnsed: Would love to not have this be a json string inside a json. Can't we flatten it into just and array of strings?

executor/batchclient/batch.go
```diff
@@ -0,0 +1,182 @@
...
+		if err != nil {
+			return "", fmt.Errorf("Failed to marshall task %s input: %s", name, err)
+		}
+		jobParams[StartingInputEnvVarName] = aws.String(string(inputStr))
+		environment = append(environment, &batch.KeyValuePair{
+			Name:  aws.String(StartingInputEnvVarName),
+			Value: aws.String(string(inputStr)),
+		})
+	} else {
+		// TODO: fix: AWS does not like empty string here
```
burnsed: 👎

handler.go
mohit: if you look at `jsonToArgs` that's exactly what we're doing. we're going from input.Data`[]interface{}` to `[]string`

resources/workflows.go
mohit: These are all just `Getters`. None of these can be set explicitly on an existing State.

resources/workflows.go
mohit: :nods: also from states language

launch/workflow-manager.yml
mohit: removed the whole `AWS_PROFILE` path. This is now handled by `ark start -l`

launch/workflow-manager.yml
mohit: https://github.com/Clever/workflow-manager/pull/2/files#diff-3254677a7917c6c01f55212f86c57fbfR6 from `template-wag`

resources/tasks.go
mohit: it's actually the containerArn from ECS. Renamed to `ContainerId` and added a comment to clarify.

resources/workflows.go
mohit: following the states language spec. don't have to at this point in the code but makes it simpler :)

mohit:
Thanks. responded to all the nits and commented on the `jsonToArgs`. Here is an example request that works:

```
curl -XPUT http://localhost:8080/jobs -d '{"workflow": {"name": "s3-to-redshift-dev", "revision": 0 }, data": ["--bucket", "REDACTED-dev", "--schema", "mongo", "--tables", "boo", "--truncate", "--config", "s3://REDACTED-dev/boo-bblah-24T21:00:00Z.yml", "--date", "2017-02-24T21:00:00Z", "--force"]}'```

burnsed:
Awesome. LGTM.

PR #6, opened 2017-04-11
Build all containers

mohit:
testing INFRA-2249 (internal)

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -16,7 +16,7 @@ test:
   - make test
 deployment:
   non-master:
-    branch: /^(?!master$)/
+    branch: /^(?!master$).*$/
     owner: Clever
     commands:
     - $HOME/ci-scripts/circleci/docker-publish $DOCKER_USER $DOCKER_PASS "$DOCKER_EMAIL" $DOCKER_ORG
--- a/circle.yml
+++ b/circle.yml
@@ -15,16 +15,18 @@ test:
   override:
   - make test
 deployment:
-  all:
-    branch: /.*/ 
+  non-master:
+    branch: /^(?!master$)/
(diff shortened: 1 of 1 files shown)


nathanleiby:
Unfortunately, I learned that if you have two deploy blocks that both match, CircleCI will (randomly?) choose one.

I think that means something like this is the way to achieve building on every branch, but only publishing/deploying on master...
https://github.com/Clever/kinesis-to-firehose/pull/9/commits/0d9d0c2e4c80be170929e7231342cfa328aea637

Another alternative:

Match branch with `/.*/`, but then use `if [ "CIRCLE_BRANCH" == "master" ]; then ... ; fi` to execute some of the lines for master branch only.

Any preference?

circle.yml
```diff
@@ -16,12 +16,16 @@ test:
...
-    branch: master
+    branch: /.*/ 
     owner: Clever
     commands:
     - $HOME/ci-scripts/circleci/docker-publish $DOCKER_USER $DOCKER_PASS "$DOCKER_EMAIL" $DOCKER_ORG
     - $HOME/ci-scripts/circleci/catapult-publish $CATAPULT_URL $CATAPULT_USER $CATAPULT_PASS $APP_NAME
-    - $HOME/ci-scripts/circleci/dapple-deploy $DAPPLE_URL $DAPPLE_USER $DAPPLE_PASS $APP_NAME
-    - $HOME/ci-scripts/circleci/npm-publish $NPM_TOKEN gen-js/
+  production:
+    branc: master
```
nathanleiby: nit: s/branc/branch

mohit:
I like what you do with kinesis-to-firehose more than the `if`

PR #7, opened 2017-04-17
use iam roles

mohit:
want to finish this up before deploying to prod


mohit:
also https://github.com/Clever/ark-config/pull/373

PR #8, opened 2017-04-20
add persistence using dynamo

rgarcia:
https://clever.atlassian.net/browse/INFRA-2244

This PR adds persistence for workflow definitions and jobs using dynamo.

Persisting `WorkflowDefinition` objects is tricky since they contain references to `State` interface types. The solution here `gob`-encodes the `WorkflowDefinition` to get around that, but could make future migrations difficult.

I added a `CreatedAt` field to both `WorkflowDefinition`s and `Job`s. I also added an `LastUpdated` field for `Job`s, since they can be mutated.

Querying for all jobs by workflow required a global secondary index on workflow name + created at time.

Smaller things:
- Fixed `succeeded` typos
- Rewired the tests in the `store` package so that they were implementation-agnostic, i.e. the same test suite is applied to `memory` and `dynamodb` stores.

@mohit I figure this PR is already big enough, so I didn't change `main()` to use dynamo. I will do a separate PR on this repo + infra repo with that change. I also will rebase the commits on this branch once review is done, since they are kind of a mess.


executor/job_manager.go
```diff
@@ -63,7 +63,7 @@ func (jm BatchJobManager) UpdateJobStatus(job *resources.Job) error {
 	} else if jobRunning {
 		job.Status = resources.Running
 	} else if jobSuccess {
-		job.Status = resources.Succeded
+		job.Status = resources.Succeeded
```
mohit: 🤦‍♂️ 🙇

mohit:
Only took a cursory look right now, digging in more in a few minutes. 

One thing that has come up recently is that we would proabably want to store Tasks in the DB outside of them just being part of a `Job`. The primary reason seems to be for looking up things like _what `TaskID`s are currently running `specific_worker`. I think for now this implementaion will work as long as we are not strongly blocking this additional feature.

Makefile
```diff
@@ -1,23 +1,26 @@
 include golang.mk
 include wag.mk
 
-.PHONY: all test build run
+.PHONY: all test build run store-test
```
mohit: `store-test` does not seem to be used.

store/dynamodb/dynamodb_store.go
```diff
@@ -0,0 +1,510 @@
+package dynamodb
```
mohit: Do you think it would be beneficial to split up the struct definitions and table setup from the `store` implementation? I'm not really sure since that does make getting a sense of the whole harder, but it's one of the complaints I hear from people working on the Catapult dynamodb implementation - that it grew into a giant file.

store/dynamodb/dynamodb_store.go
```diff
@@ -0,0 +1,510 @@
...
+}
+
+func (sk ddbJobSecondaryKeyWorkflowCreatedAt) Name() string {
+	return "workflowname-createdat"
+}
+
+func (sk ddbJobSecondaryKeyWorkflowCreatedAt) AttributeDefinitions() []*dynamodb.AttributeDefinition {
+	return []*dynamodb.AttributeDefinition{
+		{
+			AttributeName: aws.String("_gsi-wn"),
```
mohit: just for my understanding - what does the `_gsi` in `_gsi_wn`, `_gsi_ca` mean?

store/dynamodb/dynamodb_store.go
```diff
@@ -0,0 +1,510 @@
...
+	}
+
+	_, err = d.ddb.PutItem(&dynamodb.PutItemInput{
+		TableName: aws.String(d.workflowsTable()),
+		Item:      data,
+		ExpressionAttributeNames: map[string]*string{
+			"#N": aws.String("name"),
+			"#V": aws.String("version"),
+		},
+		ConditionExpression: aws.String("attribute_not_exists(#N) AND attribute_not_exists(#V)"),
```
mohit: TIL

store/dynamodb/dynamodb_store.go
```diff
@@ -0,0 +1,510 @@
...
+			if awsErr.Code() == dynamodb.ErrCodeConditionalCheckFailedException {
+				return store.NewConflict(job.ID)
+			}
+		}
+	}
+	return err
+}
+
+func (d DynamoDB) UpdateJob(job resources.Job) error {
+	job.LastUpdated = time.Now()
```
mohit: hmm this will not be reflected in the place that calls `store.UpdateJob` since I did not add a return (job, error). Or we could take in a `*job`. Suggestions?

store/dynamodb/dynamodb_store.go
```diff
@@ -0,0 +1,510 @@
...
+	if err := d.populateJob(&job); err != nil {
+		return resources.Job{}, err
+	}
+
+	return job, nil
+}
+
+// GetJobsForWorkflow returns the last 10 jobs for a workflow.
+// It uses a global secondary index on the workflow name + created time for a job.
+func (d DynamoDB) GetJobsForWorkflow(workflowName string) ([]resources.Job, error) {
```
mohit: I think it's okay to not modify this interface to support > 10 jobs for now. Just checking that you agree?

rgarcia: yeah, for now it seems fine. Eventually we probably want pagination

store/dynamodb/dynamodb_store.go
rgarcia: Yeah, I came close to changing this, but didn't want to expand the scope of the PR to change the `Store` interface. For now the workaround is to call `GetJob` after `UpdateJob` (see the tests). My gut says this function should be either

1. 
```go
UpdateJob(job resources.Job) (resources.Job, error)
```
or

2. 
```go
UpdateJob(job resources.Job) (*resources.Job, error)
```
or

3. A more specific / targeted update of a Job, e.g. `UpdateJobStatus(id, status)`.

(1) would make it parallel the `UpdateWorkflow` signature. However, returning a non-pointer type leads to an object copy in cases where you're assigning the return value to an existing variable, which isn't ideal. (2) would correct for that. I don't know enough about how the interface is used to know if (3) is doable.

store/dynamodb/dynamodb_store.go
rgarcia: yeah, I learned this while grepping for `PutItem` in our code and finding this https://github.com/Clever/unicreds/blob/8ddef5f30943fe4a9c59fa3cc1339e8cb8a54df2/ds.go#L408

store/dynamodb/dynamodb_store.go
rgarcia: global secondary index, workflowname, createdat.

store/dynamodb/dynamodb_store.go
rgarcia: yeah, i can break this up into multiple files

Makefile
rgarcia: nice catch! should be `dynamodb-test`, will fix

store/dynamodb/dynamodb_store.go
mohit: What about `UpdateJob(job *resources.Job) error`

I like the idea of (3), the only difference being that a `status` change would usually mean updated `Tasks`. 

Another reason why we would need to seperate Tasks into their own table, which we should get to in another task.

Makefile
rgarcia: 467c158

store/dynamodb/dynamodb_store.go
rgarcia: 4efdf91

rgarcia:
@mohit going to merge for now. Follow-up is to change main() to actually use this store, and configure some prod dynamo tables in terraform

PR #9, opened 2017-04-27
switch to dynamodb store

rgarcia:


mohit:
would require [launch.yml](https://github.com/Clever/workflow-manager/blob/7d2cbd7058245c32d015077d8f0cb8c9a844f04e/launch/workflow-manager.yml#L1) and ark-config changes too.

PR #10, opened 2017-05-04
add check for state.Next = nil; go1.8

mohit:


handler.go
```diff
@@ -54,12 +55,7 @@ func (wm WorkflowManager) UpdateWorkflow(ctx context.Context, input *models.Upda
 		return &models.Workflow{}, fmt.Errorf("Must define at least one state")
 	}
 
-	workflow, err := wm.store.LatestWorkflow(workflowReq.Name)
```
mohit: this was not being used.

PR #11, opened 2017-05-10
use nanoseconds for time parsing rather than sec

mohit:


executor/batchclient/batch.go
```diff
@@ -87,17 +88,18 @@ func (be BatchExecutor) Cancel(tasks []*resources.Task, reason string) []error {
 func (be BatchExecutor) jobToTaskDetail(job *batch.JobDetail) (resources.TaskDetail, error) {
 	var statusReason, containerArn string
 	var createdAt, startedAt, stoppedAt time.Time
+	msToNs := int64(1000 * 1000)
```
xavi-: Nit.  You could use `int64(time.Millisecond)`

executor/batchclient/batch_test.go
```diff
@@ -47,7 +47,7 @@ func TestJobToTaskDetail(t *testing.T) {
 	assert.NotNil(t, taskDetail)
 
 	assert.Equal(t, taskDetail.Status, resources.TaskStatusSucceeded)
-	assert.Equal(t, taskDetail.CreatedAt.Unix(), *jobDetail.CreatedAt)
-	assert.Equal(t, taskDetail.StartedAt.Unix(), *jobDetail.StartedAt)
+	assert.Equal(t, taskDetail.CreatedAt.Format(time.RFC3339Nano), "2017-03-28T00:45:46.376Z")
+	assert.Equal(t, taskDetail.StartedAt.Format(time.RFC3339Nano), "2017-03-28T00:46:48.178Z")
```
nathanleiby: worth adding a test around a nil time? will it be confusing the nil case is the unix epoch?

mohit: see line 52. Nil in this case ends up being `0001-01-01T00:00:00Z` I'm not sure there is a better way for having un assigned time.

executor/batchclient/batch.go
mohit: 🙇 fixed https://github.com/Clever/workflow-manager/pull/11/commits/986e29a3289ea271a9a126ae641370f7aae0e779#diff-f7b765fe9dbb8e27c0c94ea2b85ad0fbR91

executor/batchclient/batch_test.go
nathanleiby: Ah, sorry, i totally missed that. Seems OK

PR #12, opened 2017-05-10
Add job status polling

mohit:
add polling for task updates
* start a goroutine when creating a new job
* stop polling if the job status becomes Cancelled, Failed, Succeeded
* restarts will cause polling to stop; okay to deal with this right now
we need to eventually ECS task event listener via SNS/SQS
* don't update the DB if no task.Status has changed

### DIFF
--- a/executor/job_manager.go
+++ b/executor/job_manager.go
@@ -114,8 +114,9 @@ func (jm BatchJobManager) CreateJob(def resources.WorkflowDefinition, input []st
 
 func (jm BatchJobManager) pollUpdateStatus(job *resources.Job) {
 	for {
-		switch job.Status {
-		case resources.Cancelled, resources.Failed, resources.Succeeded:
+		if job.Status == resources.Cancelled ||
+			job.Status == resources.Failed ||
+			job.Status == resources.Succeeded {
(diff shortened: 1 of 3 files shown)


executor/job_manager.go
```diff
@@ -82,9 +104,36 @@ func (jm BatchJobManager) CreateJob(def resources.WorkflowDefinition, input []st
...
+		}
+		if err := jm.UpdateJobStatus(job); err != nil {
+			log.ErrorD("job-polling-error", logger.M{
+				"id":       job.ID,
+				"status":   job.Status,
+				"workflow": job.Workflow.Name(),
+				"error":    err.Error(),
+			})
+		}
+	}
```
rgarcia: should there be a polling interval here so that we're not calling `UpdateJobStatus` in a tight loop? i.e. a sleep inside the loop or a for loop on a ticker

mohit: 🤦‍♂️ forgot to add that to the commit

mohit: https://github.com/Clever/workflow-manager/pull/12/commits/94ed298431f3a6f27da41a8a578b9015d193876a

PR #13, opened 2017-05-19
executor: log job create, cancel, and status change

nathanleiby:
https://clever.atlassian.net/browse/INFRA-2299

Also modifies the job status change logic, so that we only save the job
to the store if its status has changed.

- [ ] Run "make generate"

### DIFF
--- a/executor/batchclient/batch_test.go
+++ b/executor/batchclient/batch_test.go
@@ -47,7 +47,7 @@ func TestJobToTaskDetail(t *testing.T) {
 	assert.NotNil(t, taskDetail)
 
 	assert.Equal(t, taskDetail.Status, resources.TaskStatusSucceeded)
-	assert.Equal(t, taskDetail.CreatedAt.Format(time.RFC3339Nano), "2017-03-28T00:45:46.376Z")
-	assert.Equal(t, taskDetail.StartedAt.Format(time.RFC3339Nano), "2017-03-28T00:46:48.178Z")
+	assert.Equal(t, taskDetail.CreatedAt.UTC().Format(time.RFC3339Nano), "2017-03-28T00:45:46.376Z")
+	assert.Equal(t, taskDetail.StartedAt.UTC().Format(time.RFC3339Nano), "2017-03-28T00:46:48.178Z")
 	assert.Equal(t, taskDetail.StoppedAt, time.Time{})
 }
--- a/executor/job_manager.go
+++ b/executor/job_manager.go
@@ -94,7 +94,7 @@ func (jm BatchJobManager) UpdateJobStatus(job *resources.Job) error {
 	// If the status was updated, log it and save to datastore
 	if previousStatus != job.Status {
 		log.InfoD("job-status-change", logger.M{
-			"job-id": job.ID,
+			"id":     job.ID,
 			"before": previousStatus,
 			"after":  job.Status,
 		})
--- a/executor/job_manager.go
+++ b/executor/job_manager.go
@@ -98,9 +98,8 @@ func (jm BatchJobManager) UpdateJobStatus(job *resources.Job) error {
 			"before": previousStatus,
 			"after":  job.Status,
 		})
-		return jm.store.UpdateJob(*job)
 	}
-	return nil
+	return jm.store.UpdateJob(*job)
 }
 
 // CreateJob can be used to create a new job for a workflow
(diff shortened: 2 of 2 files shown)


nathanleiby:
Note: In the task, it suggested we might emit a specific event called "job-finished". I don't do this directly, but I believe there should be a status change event via polling where after == `SUCCEEDED ` or `FAILED`

executor/job_manager.go
```diff
@@ -89,12 +91,22 @@ func (jm BatchJobManager) UpdateJobStatus(job *resources.Job) error {
 		job.Status = resources.Succeeded
 	}
 
-	return jm.store.UpdateJob(*job)
+	// If the status was updated, log it and save to datastore
+	if previousStatus != job.Status {
```
mohit: a task status might have changed without causing a change in the complete job status. So we might need to `UpdateJob` in the store even when the `status` remains the same. 

This is confusing right now - I have a task which will create a new table for `Tasks` which should make this more understandable.

executor/batchclient/batch_test.go
```diff
@@ -47,7 +47,7 @@ func TestJobToTaskDetail(t *testing.T) {
 	assert.NotNil(t, taskDetail)
 
 	assert.Equal(t, taskDetail.Status, resources.TaskStatusSucceeded)
-	assert.Equal(t, taskDetail.CreatedAt.Format(time.RFC3339Nano), "2017-03-28T00:45:46.376Z")
-	assert.Equal(t, taskDetail.StartedAt.Format(time.RFC3339Nano), "2017-03-28T00:46:48.178Z")
+	assert.Equal(t, taskDetail.CreatedAt.UTC().Format(time.RFC3339Nano), "2017-03-28T00:45:46.376Z")
```
mohit: 🤣 I have my VM set to UTC (and seems circleci also runs in UTC). sorries.

executor/job_manager.go
```diff
@@ -89,12 +91,22 @@ func (jm BatchJobManager) UpdateJobStatus(job *resources.Job) error {
...
+		log.InfoD("job-status-change", logger.M{
```
mohit: maybe also log `workflow-name:revision`

executor/job_manager.go
```diff
@@ -139,6 +151,11 @@ func (jm BatchJobManager) pollUpdateStatus(job *resources.Job) {
 
 func (jm BatchJobManager) CancelJob(job *resources.Job, reason string) error {
 	// don't cancel already succeeded tasks
+	log.InfoD("job-cancelled", logger.M{
```
mohit: workflow-name:revision

executor/job_manager.go
```diff
@@ -89,12 +91,22 @@ func (jm BatchJobManager) UpdateJobStatus(job *resources.Job) error {
...
+		})
+		return jm.store.UpdateJob(*job)
+	}
+	return nil
 }
 
 // CreateJob can be used to create a new job for a workflow
 func (jm BatchJobManager) CreateJob(def resources.WorkflowDefinition, input []string) (*resources.Job, error) {
 	job := resources.NewJob(def, input)
+	log.InfoD("job-created", logger.M{"id": job.ID})
```
mohit: workflow-name:revision

mohit: I'm not sure if we should not `inputs` here since they might contain sensitive information - but IP would really want it. just FYI or talking to myself here.

executor/job_manager.go
nathanleiby: Put back the "always save to store" logic, but still have this check to see if overall job status has changed: 4ea6b80

executor/job_manager.go
nathanleiby: Added workflow name and version, throughout 7e54965d327c3f0d005dda9c12cf4d7fd0c764ab

executor/job_manager.go
nathanleiby: Re: inputs - not sure. Do we have that data anywhere else?

nathanleiby:
@mohit also just a thought for later, but we might want to unify naming of Workflow `version` (internal) vs Workflow `revision` (external Swagger name for Workflow versions).

mohit:
yeah I have no idea why I ended up with that. must be that I started with `Revision` and then settled on `Version`. We should just change everything to just be `Revision` to match the client or do you have a preference between the two?

nathanleiby:
I have a slight preference for `Version`, since we use it widely already when talking about deploys. But OK with either!

PR #14, opened 2017-05-23
Add kayvee-go based log-routing

mohit:
Add log-routing to be able to set alerts on workflows. Don't really know yet if we should send the `id` as a dimension, but right now the dimensionality for it is pretty low so why not.

### DIFF
--- a/executor/job_manager.go
+++ b/executor/job_manager.go
@@ -97,8 +97,8 @@ func (jm BatchJobManager) UpdateJobStatus(job *resources.Job) error {
 			"id":               job.ID,
 			"workflow":         job.Workflow.Name(),
 			"workflow-version": job.Workflow.Version(),
-			"before":           previousStatus,
-			"after":            job.Status,
+			"previous-status":  previousStatus,
+			"status":           job.Status,
 		})
 	}
 	return jm.store.UpdateJob(*job)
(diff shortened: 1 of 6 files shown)


kvconfig.yml
```diff
@@ -0,0 +1,16 @@
...
+      series: "workflows.job-status"
+      dimensions: ["id", "workflow", "workflow-version", "after"]
+      stat_type: "counter"
+  job-polling-alerts:
+    matchers:
+      title: ["job-polling-error"]
+      output:
+        type: "alerts"
+        series: "workflows.status-poll-error"
+        dimensions: ["id", "status", "workflow"]
```
nathanleiby: nits:

- should we call these series `workflow-manager.<x>` to match the repo / app?
- would it make sense to use just one term, either "status-poll-error" (currently, the series name)  or "job-polling-error" (currently, the log title)?

kvconfig.yml
```diff
@@ -1,16 +1,19 @@
 routes:
+  # matchers for tracking Jobs/Workflows; previx workflows.
```
nathanleiby: s/previx/prefix

executor/job_manager.go
```diff
@@ -178,13 +173,23 @@ func (jm BatchJobManager) CancelJob(job *resources.Job, reason string) error {
 	errs := jm.executor.Cancel(tasks, reason)
 	jm.store.UpdateJob(*job)
 
-	if len(errs) > 0 {
-		return fmt.Errorf("%d of %d tasks were not cancelled", len(errs), len(tasks))
-	}
 	if len(errs) < len(tasks) {
```
mohit: @nathanleiby the logic here is totally confusing (what is success vs. failure). will tackle in a different PR.

PR #17, opened 2017-05-24
CreateAt Time fields for workflows, jobs, and tasks

Sayan-:
- Bug fix CreatedAt for workflows. 
- Add CreatedAt for jobs and tasks. 
- Refactor WorkflowDefinition to include CreateAtTime

- [ ] Run "make generate"

### DIFF
--- a/swagger.yml
+++ b/swagger.yml
@@ -2,7 +2,7 @@ swagger: '2.0'
 info:
   title: workflow-manager
   description: Minimal Workflow orchestrator for AWS Batch
-  version: 0.1.3
+  version: 0.1.4
   x-npm-package: workflow-manager
 schemes:
   - http
@@ -222,6 +222,7 @@ definitions:
         type: string
       createdAt:
         type: string
+        format: date-time
       startAt:
         type: string
       description:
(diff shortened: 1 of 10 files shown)


swagger.yml
```diff
@@ -236,7 +238,6 @@ definitions:
         type: string
       createdAt:
         type: string
-        format: date-time
```
mohit: did this not work?

swagger.yml
```diff
@@ -251,6 +252,8 @@ definitions:
     properties:
       id:
         type: string
+      createdAt:
```
mohit: let's add `startedAt` and `stoppedAt` also.

swagger.yml
Sayan-: I removed this so we could have control over the format of the time string instead of relying on the WAG code, in case we wanted to change the format for whatever reason.

swagger.yml
Sayan-: Will do!

Sayan-:
I've responded to your comments in line! I'm open to talking about the date-time thing. I'll add the other time fields as well

Sayan-:
Added the fields and switched to date-time!

PR #18, opened 2017-05-30
implement GetWorkflows in dynamodb_store

Sayan-:
- [ ] Run "make generate"

First pass GetWorkflows method. It uses a scan operation and filters the result after. It isn't pretty, but for the moment it doesn't require us to make changes to the DB schema.

### DIFF
--- a/store/dynamodb/dynamodb_store.go
+++ b/store/dynamodb/dynamodb_store.go
@@ -1,7 +1,6 @@
 package dynamodb
 
 import (
-	"errors"
 	"fmt"
 	"time"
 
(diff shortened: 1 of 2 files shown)


store/dynamodb/dynamodb_store.go
```diff
@@ -148,9 +147,34 @@ func (d DynamoDB) UpdateWorkflow(def resources.WorkflowDefinition) (resources.Wo
 	return d.GetWorkflow(newVersion.Name(), newVersion.Version())
 }
 
-// GetWorkflows TODO
+// GetWorkflows returns the latest version of all stored Workflows
 func (d DynamoDB) GetWorkflows() ([]resources.WorkflowDefinition, error) {
-	return nil, errors.New("Not yet implemented")
+	workflows := []resources.WorkflowDefinition{}
+	// Scan returns the entire table
+	results, err := d.ddb.Scan(&dynamodb.ScanInput{
```
Sayan-: So for the Scan operation this toggle doesn't exist, unfortunately! For the moment I'll stick to the reverse iteration

PR #19, opened 2017-05-31
update the workflows/{name} endpoint

Sayan-:
- [x] Run "make generate"

Major update to the workflows/{name} endpoint. The endpoint now behaves as follows:

- Returns an array of workflows
- Takes a latest parameter to determine what data to retrieve
- Takes a version parameter iff latest==false

### DIFF
--- a/handler.go
+++ b/handler.go
@@ -107,7 +107,7 @@ func (wm WorkflowManager) GetWorkflowVersionsByName(ctx context.Context, input *
 	return apiWorkflows, nil
 }
 
-// GetWorkflowByNameAndVersion allows fetching an existing Workflow by providing it's name
+// GetWorkflowByNameAndVersion allows fetching an existing Workflow by providing it's name and version
 func (wm WorkflowManager) GetWorkflowByNameAndVersion(ctx context.Context, input *models.GetWorkflowByNameAndVersionInput) (*models.Workflow, error) {
 	workflow, err := wm.store.GetWorkflow(input.Name, int(input.Version))
 	if err != nil {
--- a/handler.go
+++ b/handler.go
@@ -84,35 +84,36 @@ func (wm WorkflowManager) GetWorkflows(ctx context.Context) ([]models.Workflow, 
 	return apiWorkflows, nil
 }
 
-// GetWorkflowByName allows fetching an existing Workflow by providing it's name
-func (wm WorkflowManager) GetWorkflowByName(ctx context.Context, input *models.GetWorkflowByNameInput) ([]models.Workflow, error) {
-	apiWorkflows := []models.Workflow{}
+// GetWorkflowVersionsByName allows fetching an existing Workflow by providing it's name
+func (wm WorkflowManager) GetWorkflowVersionsByName(ctx context.Context, input *models.GetWorkflowVersionsByNameInput) ([]models.Workflow, error) {
 	if *input.Latest == true {
 		workflow, err := wm.store.LatestWorkflow(input.Name)
 		if err != nil {
 			return []models.Workflow{}, err
 		}
+		return []models.Workflow{*(apiWorkflowFromStore(workflow))}, nil
+	}
+
+	apiWorkflows := []models.Workflow{}
+	workflows, err := wm.store.GetWorkflowVersions(input.Name)
+	if err != nil {
+		return []models.Workflow{}, err
+	}
+	for _, workflow := range workflows {
 		apiWorkflow := apiWorkflowFromStore(workflow)
 		apiWorkflows = append(apiWorkflows, *apiWorkflow)
-	} else if input.Version != nil {
-		workflow, err := wm.store.GetWorkflow(input.Name, int(*input.Version))
-		if err != nil {
-			return []models.Workflow{}, err
-		}
-		apiWorkflow := apiWorkflowFromStore(workflow)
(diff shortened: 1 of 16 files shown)


nathanleiby:
I know we discussed this offline, but I just wanted to clarify an alternate approach. IMO, we should make the workflows endpoint be parallel to the [jobs endpoint](https://github.com/Clever/workflow-manager/blob/394533980accc402c1e3b72bd7274db4a9d66568/swagger.yml#L141)

Instead of `/workflows/{name}`, do `/workflows/{workflowId}`. This seems restful, since 

- `/workflows` = all workflows
- `/workflows/{workflowId}` = one workflow

In order to do filtering by name, latest, etc, we can query the `/workflows` endpoint with parameters. e.g.

- `/workflows?name=foo` - a list of all workflows named `foo` (could be empty)
- `/workflows?name=foo&latest=true` - a list containing the latest workflow named `foo` (or 0 items if foo dne)
- `/workflows?latest=true` - a list containing many workflows, but only the latest version of each

handler.go
```diff
@@ -85,13 +85,34 @@ func (wm WorkflowManager) GetWorkflows(ctx context.Context) ([]models.Workflow,
 }
 
 // GetWorkflowByName allows fetching an existing Workflow by providing it's name
-func (wm WorkflowManager) GetWorkflowByName(ctx context.Context, name string) (*models.Workflow, error) {
-	workflow, err := wm.store.LatestWorkflow(name)
-	if err != nil {
-		return &models.Workflow{}, err
+func (wm WorkflowManager) GetWorkflowByName(ctx context.Context, input *models.GetWorkflowByNameInput) ([]models.Workflow, error) {
```
mohit: Change this method name to be `GetWorkflowsByName` or `GetWorkflowVersionsByName`?

handler.go
```diff
@@ -85,13 +85,34 @@ func (wm WorkflowManager) GetWorkflows(ctx context.Context) ([]models.Workflow,
...
+	apiWorkflows := []models.Workflow{}
+	if *input.Latest == true {
```
mohit: what about 
```golang

if *input.Latest == true and input.Version != nil {
 		return []models.Workflow{}, "....a 400 error since this does not make sense"
}

if *input.Latest == true {
 		workflow, err := wm.store.LatestWorkflow(input.Name)
 		if err != nil {
 			return []models.Workflow{}, err
 		}
                return []models.Workflow{apiWorkflowFromStore(workflow)}, nil
}

if input.Version != nil {
 		workflow, err := wm.store.GetWorkflow(input.Name, int(*input.Version))
 		if err != nil {
 			return []models.Workflow{}, err
 		}
                return []models.Workflow{apiWorkflowFromStore(workflow)}, nil
}

workflows, err := wm.store.GetWorkflowVersions(input.Name)
if err != nil {
	return []models.Workflow{}, err
}
apiWorkflows := []models.Workflow{}
for _, workflow := range workflows {
 	apiWorkflow := apiWorkflowFromStore(workflow)
 	apiWorkflows = append(apiWorkflows, *apiWorkflow)
 }
return apiWorkflows, nil
```

swagger.yml
```diff
@@ -2,7 +2,7 @@ swagger: '2.0'
 info:
   title: workflow-manager
   description: Minimal Workflow orchestrator for AWS Batch
-  version: 0.1.4
+  version: 1.0.0
```
mohit: I think `0.2.0` would work to ensure that we don't have an auto updated client

mohit:
either solution seems fine to me.

mohit:
@nathanleiby your solution seems great to me. Sayan what do you think?

mohit:
@Sayan- @nathanleiby Here's what I think what we should do with the Workflows API to keep it consistent and simple:

### `/workflows?latest=true` 

- _returns_ a list of `Workflow` objects.
- if `latest=false` returns `ALL` Workflow objects. 
- if `latest=true` returns the highest version Workflow objects for each `name`.
- never `404` -> return empty list on no workflows found. 

### `/workflows/{name}?latest=true`

- _returns_ a list of `Workflow` objects
- if `latest=false` returns `ALL` Workflow objects with `name`
- if `latest=true` returns the highest version Workflow object for `name` (still a list).
- `404` if name not found

### `/workflows/{name}/{version}`

- _returns_ a single `Workflow` object. 
- `404` if version or name not found

Sayan-:
I think that would be great! I'll get to the changes!

Sayan-:
I'll do the /workflows?latest=true in another PR, since I want to re-work that part of the code 👍

swagger.yml
Sayan-: Roger that. I went with major since it will break backwards compatibility

handler.go
Sayan-: I've split them up per our new endpoints

nathanleiby:
@Sayan- one tip to make these commits easier to review -- you can make 2 commits, one with manual changes and the other with automated changes

For example, changes to `swagger.yml` or the `db` package would be manual changes, whereas anything in `gen-go` (basically, files outputted by `make generate`) would go in the automated changes

Sayan-:
Thanks for the tip! I'll make sure to do that from now on!

Sayan-:
Sure thing. Thanks again for the feedback!

PR #20, opened 2017-06-02
Get workflows/ endpoint update

Sayan-:
- [ ] Run "make generate"

This changes the behavior of the workflows/ endpoint to optionally take a `latest` parameter. Additionally re-imagine how we are storing and retrieving the latest versions of all workflows. We create a new table to keep track of latest versions of every workflow. Every time we save a workflow we write to the latest-workflow table and the workflow table. There are possible issues with the 2 stage write. I'm open to discussion :)

### DIFF
--- a/handler.go
+++ b/handler.go
@@ -70,18 +70,9 @@ func (wm WorkflowManager) UpdateWorkflow(ctx context.Context, input *models.Upda
 	return apiWorkflowFromStore(workflow), nil
 }
 
-// GetWorkflows fetches either:
-// 1. A list of all versions of all workflows
-// 2. A list of the latest version of each workflow
-func (wm WorkflowManager) GetWorkflows(ctx context.Context, input *models.GetWorkflowsInput) ([]models.Workflow, error) {
-	workflows := []resources.WorkflowDefinition{}
-	var err error
-
-	if *input.Latest == true {
-		workflows, err = wm.store.GetLatestWorkflows()
-	} else {
-		workflows, err = wm.store.GetWorkflows()
-	}
+// GetWorkflows retrieves a list of the latest version of each workflow
+func (wm WorkflowManager) GetWorkflows(ctx context.Context) ([]models.Workflow, error) {
+	workflows, err := wm.store.GetWorkflows()
 
 	if err != nil {
 		return []models.Workflow{}, err
--- a/handler.go
+++ b/handler.go
@@ -70,9 +70,19 @@ func (wm WorkflowManager) UpdateWorkflow(ctx context.Context, input *models.Upda
 	return apiWorkflowFromStore(workflow), nil
 }
 
-// GetWorkflows fetches all available workflows
-func (wm WorkflowManager) GetWorkflows(ctx context.Context) ([]models.Workflow, error) {
-	workflows, err := wm.store.GetWorkflows()
+// GetWorkflows fetches either:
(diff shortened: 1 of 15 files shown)


handler.go
```diff
@@ -70,9 +70,19 @@ func (wm WorkflowManager) UpdateWorkflow(ctx context.Context, input *models.Upda
 	return apiWorkflowFromStore(workflow), nil
 }
 
-// GetWorkflows fetches all available workflows
-func (wm WorkflowManager) GetWorkflows(ctx context.Context) ([]models.Workflow, error) {
-	workflows, err := wm.store.GetWorkflows()
+// GetWorkflows fetches either:
+// 1. A list of all versions of all workflows
+// 2. A list of the latest version of each workflow
```
rgarcia: It will be hard to support (1) indefinitely. Why do we still need to support it? I was expecting this change to move us exclusively to (2)

store/dynamodb/dynamodb_store.go
```diff
@@ -68,7 +88,30 @@ func (d DynamoDB) InitTables() error {
...
+			},
+		},
+		ProvisionedThroughput: &dynamodb.ProvisionedThroughput{
+			ReadCapacityUnits:  aws.Int64(1),
+			WriteCapacityUnits: aws.Int64(1),
+		},
+		TableName: aws.String(d.latestWorkflowsTable()),
+	}); err != nil {
+		return err
+	}
```
rgarcia: you'll want to update these files as well
https://github.com/Clever/infra/blob/master/config/aws/prod/dynamodb.tf#L110-L153
https://github.com/Clever/infra/blob/master/config/aws/dev/dynamodb.tf#L102-L145

store/dynamodb/dynamodb_store.go
```diff
@@ -147,39 +196,37 @@ func (d DynamoDB) UpdateWorkflow(def resources.WorkflowDefinition) (resources.Wo
...
-			continue
-		}
-		workflows = append(workflows, wf)
-		previousWorkflow = wf.NameStr
+	return d.dynamoItemsToWorkflows(results.Items)
+}
+
+// GetLatestWorkflows returns the latest version of all stored workflow definitions
+func (d DynamoDB) GetLatestWorkflows() ([]resources.WorkflowDefinition, error) {
+	// Scan returns the entire table
```
rgarcia: nit, somewhat obvious comment

store/dynamodb/dynamodb_store.go
```diff
@@ -239,6 +280,7 @@ func (d DynamoDB) GetWorkflow(name string, version int) (resources.WorkflowDefin
 
 // LatestWorkflow gets the latest version of a workflow definition.
 func (d DynamoDB) LatestWorkflow(name string) (resources.WorkflowDefinition, error) {
+	// TODO we should change this to use the latest-workflows table with a projection expressions
```
rgarcia: explain? From what I can tell it should be as simple as changing `TableName` in the query below to the latest table

handler.go
mohit: +1 on only supporting 2 in the long run. I did not want to overrotate on that right now unless it added additional implementation burden :)

Sayan-: Completely agree - I can't remember the exact use case we discussed of offline where we would want to fetch everything. I can go ahead and remove (1) this pass instead of having to prune it later on

PR #22, opened 2017-06-29
Dynamic jobdefinition lookup

mohit:
- providing a namespace to the CreateJob endpoint performs a lookup in
the StateResource table to lookup the BatchJobDefinitionARN.

- leaving the namespace blank (== "") mirrors the current behavior where
it is expected that the Workflow.State.Resource is set to a valid ARN.

- Another minor change here is to add StateResource to the `Task` model.
This will allow us to lookup the JobDefinitionARN used to create the
task.

----------------

- [ ] Run "make generate"

### DIFF
--- a/executor/job_manager.go
+++ b/executor/job_manager.go
@@ -14,7 +14,7 @@ var log = logger.New("workflow-manager")
 
 // JobManager in the interface for creating, stopping and checking status for Jobs (workflow runs)
 type JobManager interface {
-	CreateJob(def resources.WorkflowDefinition, input []string) (*resources.Job, error)
+	CreateJob(def resources.WorkflowDefinition, input []string, namspace string) (*resources.Job, error)
 	CancelJob(job *resources.Job, reason string) error
 	UpdateJobStatus(job *resources.Job) error
 }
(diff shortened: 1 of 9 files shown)


executor/job_manager.go
```diff
@@ -14,7 +14,7 @@ var log = logger.New("workflow-manager")
 
 // JobManager in the interface for creating, stopping and checking status for Jobs (workflow runs)
 type JobManager interface {
-	CreateJob(def resources.WorkflowDefinition, input []string) (*resources.Job, error)
+	CreateJob(def resources.WorkflowDefinition, input []string, namspace string) (*resources.Job, error)
```
nathanleiby: nit: s/namspace/namespace

executor/job_manager.go
```diff
@@ -219,16 +226,55 @@ func (jm BatchJobManager) scheduleTasks(job *resources.Job, input []string) erro
...
-		task := resources.NewTask(taskID, taskName, state.Name(), taskInput)
+		task := resources.NewTask(taskID, taskName, state.Name(), stateResources[state.Name()], taskInput)
 		tasks[state.Name()] = task
 		job.AddTask(task)
 	}
 
 	return nil
 }
+
+// getStateResources fetches JobDefinition URIs for each state 8
```
nathanleiby: nit: remove `8`

executor/job_manager.go
```diff
@@ -219,16 +226,55 @@ func (jm BatchJobManager) scheduleTasks(job *resources.Job, input []string) erro
...
+// from store.StateResource if namespace is set. If namespace is NOT
+// defined then StateResource objects are created with URI = state.Resource
+//
+// This behavior allows to shortcircuit use of the StateResource database and provide
+// Resource URIs directly in the WorkflowDefinition
```
nathanleiby: Do we want this mixed mode to co-exist in the long run, or is it just useful for migration?

mohit: There is some benefit in providing a way to completely short-circuit the use of `StateResources`, however I don't think I plan to keep this in the long run. It's useful right now for migration and testing as we built stuff out.

mohit:
Tested end-to-end and the model changes are backwards compatible. i.e both versions of the model can be read by both master and this branch. The only different is that old `Job` objects have `StateResource=nil`. Since we're not using that yet anyways, merging.

PR #24, opened 2017-07-18
Support multiple queues

nathanleiby:
- [x] Run "make generate"

Previously, the batch executor was tied to exactly one queue. 

In order to allow targeting other queues, we add `queue` as an argument to the SubmitJob (`POST /jobs`) endpoint.

The workflow manager's configuration now accepts a default queue (exactly 1), and custom queues (0 or more).

- default queue - if an empty string is used as the value of queue (`""`) in submit job, we'll always fallback to this queue
- custom queues - this is a map from `name => queue`.
  - we don't use queues name directly, in order to allow a layer of abstraction between clients and the workflow-manager
  - for example, clients might submit to a "queue" called `high-priority` which in AWS land is called `aws-batch-queue-foo-123-fancyname`

--

Also, updates batch executor tests to mock the AWS Batch client, using `gomock`.


nathanleiby:
Targeting more than one queue unblocks one use cases for us:

- dev vs prod queues

It also allows for greater flexibility in the future, e.g. multiple prod queues with different priorities.

executor/batchclient/batch.go
```diff
@@ -162,7 +164,17 @@ func (be BatchExecutor) taskStatus(job *batch.JobDetail) resources.TaskStatus {
 }
 
 // SubmitJob queues a task using the AWS Batch API client and returns the taskID
-func (be BatchExecutor) SubmitJob(name string, definition string, dependencies []string, input []string) (string, error) {
+func (be BatchExecutor) SubmitJob(name string, definition string, dependencies []string, input []string, queue string) (string, error) {
+	var ok bool
```
mohit: we should just abstract this as an internal function on `be`, since I'm sure we will be using this more in the future :p

nathanleiby: refactored 6b87f08

PR #25, opened 2017-07-18
allow namespace to be "" and use `default` in task

mohit:
this allows us to keep the {namespace--app} format without making
namespace a required arg


nathanleiby:
lgtm

PR #26, opened 2017-07-18
INFRA-2390: also log a numerical value for task/job failures

mohit:
this makes it easy for aggregating the status of jobs for all workflows
and for task errors across jobs without the need for excessive
heartbeats.

### DIFF
--- a/executor/job_manager.go
+++ b/executor/job_manager.go
@@ -69,6 +69,13 @@ func (jm BatchJobManager) UpdateJobStatus(job *resources.Job) error {
 	jobRunning := false
 	jobFailed := false
 	for _, task := range job.Tasks {
+		log.InfoD("task-status", logger.M{
+			"id":         job.ID,
+			"workflow":   job.Workflow.Name(),
+			"state":      task.State,
+			"status":     task.Status,
+			"status_num": task.StatusToInt(),
+		})
 		if task.Status != resources.TaskStatusSucceeded {
 			// all tasks should be successful for job success
 			jobSuccess = false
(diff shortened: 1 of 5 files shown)


executor/job_manager.go
```diff
@@ -112,7 +120,8 @@ func (jm BatchJobManager) CreateJob(def resources.WorkflowDefinition, input []st
 		"workflow":         job.Workflow.Name(),
 		"workflow-version": job.Workflow.Version(),
 		"previous-status":  "",
-		"status":           job.Status, // CREATED
+		"status":           job.Status,        // QUEUED
+		"status_num":       job.StatusToInt(), // 0
```
mohit: Any suggestions for a better name for this field are much appreciated. I went through `error`, `failed` and `exit_code`

executor/job_manager.go
```diff
@@ -190,6 +199,7 @@ func (jm BatchJobManager) CancelJob(job *resources.Job, reason string) error {
 			"previous-status":  previousStatus,
 			"status":           job.Status,
 			"reason":           reason,
+			"status_num":       job.StatusToInt(), // non-zero (-1)
```
nathanleiby: since we have this in 3 places worth pulling out into a common method (like we do for logging app lifecycle events in catapult)?

nathanleiby:
any reason for negative status codes?

executor/job_manager.go
mohit: started by adding a `log_helpers.go` file to the `executor` package. This abstracts the job and task status loggers. We can grow this, and move it into another package as it grows: https://github.com/Clever/workflow-manager/pull/26/commits/fd3acbea8f86f06453772f62c51b68bfbf77a0b4

mohit:
here is the current heuristic I'm using for the numbers associated with errors:

* `0`: things are in flight or completed successfully
* `+ve`: an error/failure in the job/task - akin to a 5XX.
* `-ve`: an external event (not intrinsic to the task/job) caused an unsuccessful run - user or a dependency had a failure. 

For our alerting system this means we can easily Exclude +ve or -ve values when setting up alerts and compare them to the total number of tasks.

PR #27, opened 2017-07-19
Add delete state resource

mohit:
missed a spot. 

- [x] Run "make generate"


store/dynamodb/dynamodb_store.go
```diff
@@ -357,6 +357,39 @@ func (d DynamoDB) GetStateResource(name, namespace string) (resources.StateResou
 	return stateResource, nil
 }
 
+// DeleteStateResource removes an existing StateResource matching the name and namespace
+func (d DynamoDB) DeleteStateResource(name, namespace string) error {
+	// TODO: maybe we want to mark for deletion instead?
```
rgarcia: what's the risk of deleting instead of marking?

mohit: Mostly it would allow us to have rollbacks that just point to the older version of the JobDefinitionARN rather than always creating a new one. I'm not sure right now that is the right direction, so just deleting works. 

I'll add the above as a comment ^ to the catapult PR I'm working on which uses this function.

PR #30, opened 2017-08-09
JobInput.Data should be []string not []interface{}

mohit:
We have standardized on using `[]string` as the data input for new Jobs but the `swagger.yml` did not specify `type: string`.

Followup: need to make sure none of the existing consumers break due to this change. 

Test with new client before merging 
- [ ] clever-cron
- [ ] mp-trigger
- [ ] ark

- [ ] Run "make generate"

### DIFF
--- a/gen-go/models/job_input.go
+++ b/gen-go/models/job_input.go
@@ -15,7 +15,7 @@ import (
 type JobInput struct {
 
 	// data
-	Data []interface{} `json:"data"`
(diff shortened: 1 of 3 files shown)


nathanleiby:
@mohit is this change still needed?

mohit:
I'll look into this today.

mohit:
Closing since #41 adds this

PR #31, opened 2017-08-15
Fix job status updates to handle cancels and tasks that no longer exists

burnsed:
Fix:
-  `cancel` to set the job status upfront.
-  `updateJobStatus` to not reset the job status once it is canceled
-  `taskStatus` prefix check for user aborted

Fixed up unittests and tested locally

- [x] Run "make generate"


burnsed:
cc @mohit

executor/batchclient/batch.go
```diff
@@ -149,7 +149,7 @@ func (be BatchExecutor) taskStatus(job *batch.JobDetail) resources.TaskStatus {
 		return resources.TaskStatusSucceeded
 	case batch.JobStatusFailed:
 		if job.StartedAt == nil {
-			if strings.HasPrefix(string(resources.TaskStatusUserAborted), statusReason) {
+			if strings.HasPrefix(statusReason, string(resources.TaskStatusUserAborted)) {
```
mohit: derp

executor/job_manager_test.go
```diff
@@ -112,6 +112,7 @@ func TestUpdateJobStatus(t *testing.T) {
 		mockClient.tasks[task.ID].SetStatus(resources.TaskStatusFailed)
```
mohit: Should we add additional tests for the Cancel cases that were fixed in this PR.

swagger.yml
```diff
@@ -7,7 +7,7 @@ info:
   # 2. merge the new clients, and, after merging, tag the commit with the version:
   #    git tag -a vX.Y.Z -m "vX.Y.Z"
   #    git push origin --tags
-  version: 0.3.3
+  version: 0.3.4
```
mohit: should we really bump the version when there are no API changes?

burnsed: no idea. I just saw the checklist in the PR template :grin:

executor/job_manager_test.go
burnsed: good idea. let me add some...

swagger.yml
mohit: hmm not sure either way. maybe @xavi- has an opinion here

xavi-: Don't have strong opinions, but it seems like a good idea.  If all wag clients do this, I can imagine a world where services are more strict about which client they'll accept requests from.

executor/job_manager_test.go
burnsed: Added unittests for the 3 cases I fixed

resources/tasks.go
```diff
@@ -48,10 +48,13 @@ func NewTask(id, name, state string, stateResource StateResource, input []string
 	}
 }
 
-// IsComplete can be used check if a task is completed
-// true if failed or succeded
-func (t *Task) IsComplete() bool {
-	return (t.Status == TaskStatusFailed || t.Status == TaskStatusSucceeded)
+// IsFinal can be used check if a task's state is expected to change
+// true if the task is in a final state; false if its status might still change
+func (t *Task) IsFinal() bool {
```
xavi-: Nit. I feel `IsDone` a bit more of a familiar naming scheme.

PR #32, opened 2017-08-21
Use TaskArn rather than Container Instance ARN

alsmola:
See https://clever.atlassian.net/browse/INFRA-2439

This change simply replaces the `ContainerInstanceArn` with the `TaskArn` for a job's container. I'm not convinced this value will be present, so I'd appreciate advice about how to test before deploying.


mohit:
We can run a job using a dev `workflow-manager` (happy to pair on setup since it's a little manual right now).

mohit:
Lgtm! Tested by running a job in dev

PR #33, opened 2017-08-23
Change name of ContainerID to ECSTaskID

alsmola:
The ContainerID value is actually the ECSTaskID. This changes the name - the value was changed in https://github.com/Clever/workflow-manager/pull/32.

- [x] Run "make generate"


mohit:
This is a breaking API change that will require changing the consumers of workflow manager. Can we create a ticket instead and look into the implications a little more

alsmola:
Filed https://clever.atlassian.net/browse/INFRA-2483

PR #34, opened 2017-08-25
add retries

rgarcia:
https://clever.atlassian.net/browse/INFRA-2431

This adds support for specifying a "retry" section in a workflow definition state, e.g.:

```json
{
    "name": "batchcli-test:local",
    "startAt": "echo-1",
    "states": [
        {
            "name": "echo-1",
            "resource": "raf--batchcli:1",
            "type": "Task",
            "retry": [{
                "errorEquals": ["States.ALL"],
                "maxAttempts": 2
            }],
            "end": true
        }
    ]
}
```

The syntax (`errorEquals` + `maxAttempts` + `states.ALL`) is taken from the [states language spec](https://states-language.net/spec.html#retrying-after-error). It is a subset of what is supported there, namely you have to specify `States.ALL` (we don't support error state matching) and you must use `maxAttempts` (since AWS Batch only supports this kind of retry strategy).

The changes to the data model are all summarized in `swagger.yml`. The `State` model has expanded to have the `Retry` section. In addition, I've added an `Attempts` field to the `Task` model, which is a passthrough for the data returned from the AWS DescribeJobs API. It is stored when we receive updated job status. It is not surfaced anywhere right now (unless you count the /jobs endpoint).


mohit:
Still working through the changes but just wanted to mention that we would need to update debs in `ark` so that `ark wfbeta status` does not break

executor/batchclient/batch_test.go
```diff
@@ -86,19 +86,19 @@ func TestSubmitJobToCustomQueue(t *testing.T) {
 	mockClient.EXPECT().SubmitJob(gomock.Any()).Return(&batch.SubmitJobOutput{
 		JobId: aws.String("job-id"),
 	}, nil)
-	out, err := be.SubmitJob(name, definition, dependencies, input, "")
```
mohit: We should add tests for setting the retry if possible

resources/kitchensink.go
```diff
@@ -18,20 +19,23 @@ func KitchenSinkWorkflow(t *testing.T) WorkflowDefinition {
 			"fake-resource-1",
 			[]string{},
 			false,
+			[]*models.Retrier{},
```
mohit: add a state here with retries since we want to have the `KitchenSink` contain all features.

resources/tasks.go
```diff
@@ -1,6 +1,10 @@
 package resources
 
-import "time"
+import (
+	"time"
+
+	"github.com/Clever/workflow-manager/gen-go/models"
```
mohit: I had attempted to create a separation between the `gen-go` models and resources but since I can't remember a clear reason why 🤷‍♂️

swagger.yml
```diff
@@ -318,6 +318,29 @@ definitions:
         type: string
       end:
         type: boolean
+      retry:
```
mohit: Just for my understanding - it's optional by default right?

swagger.yml
```diff
@@ -382,6 +405,28 @@ definitions:
...
+  TaskAttempt:
+    type: object
+    properties:
+      startedAt:
+        type: string
+        format: date-time
+      stoppedAt:
+        type: string
+        format: date-time
+      containerInstanceARN:
```
mohit: We probably don't need `containerInstanceARN` to be part of the API.

swagger.yml
```diff
@@ -382,6 +405,28 @@ definitions:
...
+        type: string
+      taskARN:
```
mohit: Also change `Task.Container` to be `Task.TaskARN` since we're going to have to update ark for this change?

executor/job_manager.go
```diff
@@ -222,7 +224,18 @@ func (jm BatchJobManager) scheduleTasks(job *resources.Job,
...
 			taskInput = input
 		}
-		taskID, err = jm.executor.SubmitJob(taskName, taskDefinition, deps, taskInput, queue)
+
+		// determine if the state has a retry strategy
+		// currently we just support a retry strategy corresponding to a number of attempts
+		var attempts int64
+		if len(state.Retry()) == 1 {
+			retrier := state.Retry()[0]
+			if retrier.MaxAttempts != nil {
```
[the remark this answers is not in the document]
rgarcia: to be safe I generally put nil checks before dereferencing pointers

resources/kitchensink.go
rgarcia: 8c2ff19

resources/tasks.go
rgarcia: I like using the generated models since it makes it easier to see when you've broken things. E.g. AFAICT I haven't broken anything in the APIs (I've only added fields), so I don't think `ark` changes are required.

swagger.yml
rgarcia: yeah

swagger.yml
rgarcia: I thought you mentioned it being useful to know what instance something ran on for debugging?

swagger.yml
rgarcia: I don't think ark changes are required, I was able to use it w/o modification

rgarcia: Maybe add the Task.Container -> Task.TaskARN to refactor task?

rgarcia:
@mohit as written none of these changes require ark modifications, but if we start renaming existing things then we would need to change ark

swagger.yml
mohit: Will do. Thanks!

swagger.yml
mohit: 🤦‍♂️ agree I have found this useful for that reason in the past.

PR #35, opened 2017-08-31
return ordered states from API

rgarcia:
https://clever.atlassian.net/browse/INFRA-2475

ngbbs455f5e7aafb04


nathanleiby:
LGTM. triggered a rebuild in circle CI since there was an issue in `docker-publish` step

PR #36, opened 2017-09-08
Refactor: Rename various Types, update API routes (breaking changes)

nathanleiby:
https://clever.atlassian.net/browse/INFRA-2483

### UPDATED (Sept 8 3:30pm PT)

Here, we do several major renames

1. Workflow => WorkflowDefinition
1. Job => Workflow
1. Task => Job

Breaking changes include:

- API paths (e.g. /workflows -> /workflow-definitions) (54d2224)

- Renames in `swagger.yml` cause the field names for some objects, hence their JSON marshalled representation will be different when storing into Dynamo
- updated hard coded strings for DynamoDB table/field names (e0c1f9483d8f03f46b81ae0fa3793e3fe63037fe)

Followup work:

- rename files appropriately (didn't change here, in order to make diffs easier to read)
- update text in logs and related kvconfig.yml for log routing

- [x] Run "make generate"


nathanleiby:
The goal is to do several small PRs like this to rename things bit-by-bit, which aren't breaking changes.

I'd like to merge these in fairly quickly, so I'm not fighting against a moving target on master.

--

Eventually, we'll also need to apply two breaking changes:

1. updating the API paths (e.g. `/workflows` -> `/workflow-definitions`)
2. renaming the DynamoDB tables

controller.go
```diff
@@ -0,0 +1,340 @@
+package main
```
mohit: why change from handler.go which is the standard in most of our go services to controller.go?

controller.go
```diff
@@ -0,0 +1,340 @@
...
+
+	"github.com/Clever/workflow-manager/executor"
+	"github.com/Clever/workflow-manager/gen-go/models"
+	"github.com/Clever/workflow-manager/resources"
+	"github.com/Clever/workflow-manager/store"
+	"github.com/go-openapi/strfmt"
+)
+
+// controller implements the wag server
+type controller struct {
```
mohit: afaik type names should be `TitleCase`?

swagger.yml
```diff
@@ -431,16 +435,16 @@ definitions:
   JobInput:
     type: object
     properties:
-      workflow:
```
mohit: This breaks the API I think

swagger.yml
```diff
@@ -431,16 +435,16 @@ definitions:
...
-        $ref: '#/definitions/WorkflowRef'
+      workflowDefinition:
```
mohit: breaks the api

controller.go
nathanleiby: Although the file was called handler.go, the actual struct it had inside was called WorkflowManager.

ex

```golang
func (wm WorkflowManager) HealthCheck(ctx context.Context) error {
```

I first renamed this struct to `Handler`. However, I then renamed it to `controller` (and renamed the file too), since `Controller` is the interface it implements in wag.

I agree it's an un-related change.

controller.go
nathanleiby: Ah, I was thinking that the case meant whether or not something was public or private.. In this case, it doesn't need to be public, so I left it lowercase. But I realize the methods on it are public (since they implement the `Controller` interface), so... making it `c` => `C` sounds good to me

swagger.yml
nathanleiby: good point! thanks for the catch. I didn't have this here originally but did some more refactoring which broke this.

I'll try to just make all the changes needed in one big breaking PR today.

controller.go
nathanleiby: For an example of a different setup..

In app service, `main.go` function just reads config and starts the server, and the controller is implemented in the `controller` package.

https://github.com/Clever/app-service/tree/master/controller

I think I actually prefer that approach as things get bigger, since it makes it easier for each of your different base routes to be split into different files within the controller package.

nathanleiby:
@mohit I ended up going with the "breaking changes" approach we discussed offline. I've updated the PR description accordingly. 

I now plan to rollout / test e2e in a dev environment.

controller.go
nathanleiby: renamed controller to handler (also changed back the filename) 3d0c7c5

PR #37, opened 2017-09-12
refactor: rename files to match renamed types

nathanleiby:
Just renames some files. 

The ones where you see a diff are where we had a naming collision. For example, `git mv jobs.go workflows.go` appears to be a diff since workflows.go existed already (but it contained `WorkflowDefinitions` after https://github.com/Clever/workflow-manager/pull/36)


nathanleiby:
We didn't rename the files in the original refactor, since that made it easier to read and reason about actual code changes.

PR #38, opened 2017-09-12
circle: enable dapple 'confirm-then-deploy'

nathanleiby:
Hooks workflow manager up to our CD system. Will send a prompt to Slack
to confirm deployments.


mohit:
Yay!

PR #42, opened 2017-09-14
Split out config for the various dynamo tables.

kofi-clever:
This is so we can read from the prod definitions and state resources tables in local/staging, but continue to write to the staging workflows table.

Config added in https://github.com/Clever/ark-config/pull/604

- [N/A] Run "make generate"

### DIFF
--- a/main.go
+++ b/main.go
@@ -22,11 +22,13 @@ import (
 
 // Config contains the configuration for the workflow-manager app
 type Config struct {
-	AWSRegion         string
-	DefaultBatchQueue string
-	CustomBatchQueues map[string]string
-	DynamoRegion      string
-	DynamoPrefix      string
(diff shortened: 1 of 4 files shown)


mohit:
in description: `s/write to staging workflows\/state resources/write to staging workflows/`

main.go
```diff
@@ -75,35 +81,29 @@ func awsSession(c Config) *session.Session {
...
 		log.Fatal(err)
 	}
 
-	if dynamoPrefix == "" {
-		dynamoPrefix = "workflow-manager-test"
-	}
-
 	return Config{
-		AWSRegion:         region,
-		DefaultBatchQueue: defaultQueue,
```
mohit: 💯

PR #43, opened 2017-09-15
remove polling loop on individual workflows

rgarcia:
https://clever.atlassian.net/browse/INFRA-2300

Before this change, workflow-manager would initiate a goroutine on workflow submission that monitored the workflow for updates. This introduced state to workflow-manager that is lost on restart, similar to what we have today with gearman-admin.

This PR removes the polling loop on individual workflows and replaces it with a polling loop that is running constantly looking for workflows in a pending state. When it finds one, it locks it, and then attempts to perform an update on it.

There are roughly three parts to this change, broken down into three commits:
1) add a `GetPendingWorkflows` method to the store interface that returns workflows that are in a queued or running state.
2) add a `LockWorkflow` and `UnlockWorkflow` that lets you lock a workflow for updating, and unlock it when you're done.
3) replace the polling loop spawned on worfklow creation with a polling loop spawned on startup. It uses `GetPendingWorkflows` to find a workflow to check, then uses `LockWorkflow` to lock it, and then uses the existing method for updating a workflow.

I've added tests to the `store` package, and manually tested the polling loop in a couple of ways locally:
- Running 2 instances of workflow-manager locally, submitting a workflow, seeing its status proceed from queued -> running -> complete.
- Running 2 instances of workflow-manager locally, submitting a workflow, killing instances of workflow-manager, re-spawning them, seeing it pick up on the pending workflow and update its status.

### DIFF
--- a/executor/workflow_manager.go
+++ b/executor/workflow_manager.go
@@ -184,7 +184,8 @@ func (wm BatchWorkflowManager) checkPendingWorkflows() error {
 		wm.store.UnlockWorkflow(wfLockedID)
 	}()
 
-	if wf, err := wm.store.GetWorkflowByID(wfLockedID); err != nil {
+	wf, err := wm.store.GetWorkflowByID(wfLockedID)
+	if err != nil {
 		return err
 	}
 
--- a/executor/workflow_manager.go
+++ b/executor/workflow_manager.go
@@ -174,23 +174,21 @@ func (wm BatchWorkflowManager) checkPendingWorkflows() error {
 	}
 
 	if wfLockedID == "" {
-		log.InfoD("check-pending-workflows-noop", logger.M{})
+		log.InfoD("pending-workflows-noop", logger.M{"pending": len(wfIDs)})
 		return nil
 	}
 
-	log.InfoD("check-pending-workflow-locked", logger.M{"id": wfLockedID})
+	log.InfoD("pending-workflows-locked", logger.M{"id": wfLockedID})
 	defer func() {
-		log.InfoD("check-pending-workflow-unlocked", logger.M{"id": wfLockedID})
+		log.InfoD("pending-workflows-unlocked", logger.M{"id": wfLockedID})
 		wm.store.UnlockWorkflow(wfLockedID)
 	}()
 
 	if wf, err := wm.store.GetWorkflowByID(wfLockedID); err != nil {
 		return err
-	} else if err := wm.UpdateWorkflowStatus(&wf); err != nil {
-		return err
 	}
 
-	return nil
+	return wm.UpdateWorkflowStatus(&wf)
 }
 
 func (wm BatchWorkflowManager) CancelWorkflow(workflow *resources.Workflow, reason string) error {
--- a/executor/workflow_manager.go
+++ b/executor/workflow_manager.go
@@ -154,6 +154,18 @@ func (wm BatchWorkflowManager) PollForPendingWorkflows(ctx context.Context) {
 	}
 }
 
+func (wm BatchWorkflowManager) lockAvailableWorkflow(workflowIDs []string) (string, error) {
+	for _, wfID := range workflowIDs {
+		if err := wm.store.LockWorkflow(wfID); err == nil {
+			return wfID, nil
+		} else if err != store.ErrWorkflowLocked {
+			// an error reading from the Store
+			return "", err
+		}
+	}
+	return "", nil
+}
+
 func (wm BatchWorkflowManager) checkPendingWorkflows() error {
 	wfIDs, err := wm.store.GetPendingWorkflowIDs()
 	if err != nil {
@@ -161,18 +173,10 @@ func (wm BatchWorkflowManager) checkPendingWorkflows() error {
 	}
 
 	// attempt to lock one of the workflows for updating
-	var wfLockedID string
-	for _, wfID := range wfIDs {
-		if err := wm.store.LockWorkflow(wfID); err != nil {
-			if err == store.ErrWorkflowLocked {
-				continue
-			}
-			return err
-		}
-		wfLockedID = wfID
-		break
+	wfLockedID, err := wm.lockAvailableWorkflow(wfIDs)
+	if err != nil {
+		return err
 	}
-
 	if wfLockedID == "" {
 		log.InfoD("pending-workflows-noop", logger.M{"pending": len(wfIDs)})
 		return nil
--- a/executor/workflow_manager.go
+++ b/executor/workflow_manager.go
@@ -1,6 +1,7 @@
 package executor
 
 import (
+	"context"
 	"fmt"
 	"time"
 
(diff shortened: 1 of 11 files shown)


mohit:
cc: @kofi-clever just FYI

store/dynamodb/dynamodb_store.go
```diff
@@ -520,3 +540,47 @@ func (d DynamoDB) GetWorkflows(workflowName string) ([]resources.Workflow, error
...
+			return nil, err
+		}
+		query.TableName = aws.String(d.workflowsTable())
+		query.Limit = aws.Int64(5)
+		res, err := d.ddb.Query(query)
+		if err != nil {
+			return nil, err
+		}
+
+		for _, item := range res.Items {
```
mohit: just for my understanding: these would already be ordered by the last updated first?

glide.yaml
```diff
@@ -27,3 +27,6 @@ import:
   version: master
   subpackages:
   - gomock
+- package: github.com/zencoder/ddbsync
+  version: attempt-lock
```
mohit: should we move this to master in our fork?

store/dynamodb/dynamodb_store.go
```diff
@@ -26,10 +28,17 @@ type TableConfig struct {
...
+	d := DynamoDB{
 		ddb:         ddb,
 		tableConfig: tableConfig,
 	}
+	d.lockDB = ddbsync.NewDatabaseFromDDBAPI(ddb, d.locksTable())
+	return d
+}
+
+// locksTable returns the name of the table that stores the locks on workflows.
+func (d DynamoDB) locksTable() string {
```
mohit: This is probably not the best place to write it but just wanted to highlight a use case.

If there are thousands of Workflows active for a specific Workflow (seems like this is normal for multiverse), would we get into a state where we would need paging through `ddbWorkflowStatusLastUpdated` since all of the Workflows returned by the first page would already be locked?

Maybe this is not a concern, I'm just not sure I understand concretely now how that is ensured.

executor/workflow_manager.go
```diff
@@ -132,33 +133,64 @@ func (wm BatchWorkflowManager) CreateWorkflow(def resources.WorkflowDefinition,
...
+}
+
+func (wm BatchWorkflowManager) checkPendingWorkflows() error {
+	wfIDs, err := wm.store.GetPendingWorkflowIDs()
+	if err != nil {
+		return err
+	}
+
+	// attempt to lock one of the workflows for updating
+	var wfLockedID string
```
mohit: if we move this loop out into it's own function e.g. `lockAvailableWorkflow(workflowIds) (string, err)` we could get rid of the `continue` and `break` which make this loop harder to follow.

executor/workflow_manager.go
```diff
@@ -132,33 +133,64 @@ func (wm BatchWorkflowManager) CreateWorkflow(def resources.WorkflowDefinition,
...
+			}
+			return err
 		}
-		time.Sleep(time.Minute)
+		wfLockedID = wfID
+		break
 	}
+
+	if wfLockedID == "" {
+		log.InfoD("check-pending-workflows-noop", logger.M{})
```
mohit: _nit_: we can call all of these `pending-workflows` and add `state=none-found`, `state=locked`, `state=unlocked` to make it easy to follow the path of this function in kibana, or setup alerts in signalfx.

executor/workflow_manager.go
```diff
@@ -132,33 +133,64 @@ func (wm BatchWorkflowManager) CreateWorkflow(def resources.WorkflowDefinition,
...
+
+	log.InfoD("check-pending-workflow-locked", logger.M{"id": wfLockedID})
+	defer func() {
+		log.InfoD("check-pending-workflow-unlocked", logger.M{"id": wfLockedID})
+		wm.store.UnlockWorkflow(wfLockedID)
+	}()
+
+	if wf, err := wm.store.GetWorkflowByID(wfLockedID); err != nil {
+		return err
+	} else if err := wm.UpdateWorkflowStatus(&wf); err != nil {
```
mohit: nit: this could just be a 

```golang
return wm.UpdateWorkflowStatus(&wf)
```

since you have a `return err` in the `if` above

main.go
```diff
@@ -57,13 +58,15 @@ func main() {
...
-
+	wfm := executor.NewBatchWorkflowManager(batch, db)
 	h := Handler{
 		store:   db,
-		manager: executor.NewBatchWorkflowManager(batch, db),
+		manager: wfm,
 	}
 	s := server.New(h, *addr)
 
+	go wfm.PollForPendingWorkflows(context.Background())
```
mohit: (I don't know go) do we need an explicit defer to make sure the context is stopped when we get a SIGTERM?

store/dynamodb/dynamodb_store.go
rgarcia: Definitely a valid concern that I think will become easier to reason about as we scale up the throughput of workflows. 

One benefit of using step functions would be not having to push workflows from state to state. In this world, the loop on pending workflows would simply be a record-keeping loop to make sure the records we have in dynamo match what's in AWS.

store/dynamodb/dynamodb_store.go
```diff
@@ -520,3 +540,47 @@ func (d DynamoDB) GetWorkflows(workflowName string) ([]resources.Workflow, error
...
+func (b byLastUpdatedTime) Swap(i, j int)      { b[i], b[j] = b[j], b[i] }
+func (b byLastUpdatedTime) Less(i, j int) bool { return b[i].LastUpdated.Before(b[j].LastUpdated) }
+
+// GetPendingWorkflowIDs gets workflows that are either Queued or Running.
+// It uses a global secondary index on status and last updated time in order to return
+// workflows ordered by their last updated time. Workflows with the oldest last updated
+// time are returned first.
+func (d DynamoDB) GetPendingWorkflowIDs() ([]string, error) {
+	var pendingWorkflows []resources.Workflow
+	for _, statusToQuery := range []resources.WorkflowStatus{resources.Queued, resources.Running} {
```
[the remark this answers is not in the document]
rgarcia: I'm tempted to make a note of it and just leave it be as an edge case we can worry about in the future...

store/dynamodb/dynamodb_store.go
rgarcia: yeah

executor/workflow_manager.go
rgarcia: AFAICT the loop would be identical whether it's here or in a function like that?

executor/workflow_manager.go
rgarcia: 3375b20

executor/workflow_manager.go
rgarcia: 3375b20

executor/workflow_manager.go
mohit: maybe I am not understanding the function very well. I was thinking something like

```golang

func lockAvailableWorkflow(workflowIds) (string, err) {
  for _, wfID := range wfIDs {
	if err := wm.store.LockWorkflow(wfID); err == nil {
	  return wfID, nil
	}
	if err != store.ErrWorkflowLocked {
          // an error reading from the Store
	  return err
	}
  }

  return "", nil
}
```

store/dynamodb/dynamodb_store.go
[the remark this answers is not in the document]
mohit: 👍

executor/workflow_manager.go
```diff
@@ -132,33 +133,62 @@ func (wm BatchWorkflowManager) CreateWorkflow(def resources.WorkflowDefinition,
...
+
+	if wfLockedID == "" {
+		log.InfoD("pending-workflows-noop", logger.M{"pending": len(wfIDs)})
+		return nil
 	}
+
+	log.InfoD("pending-workflows-locked", logger.M{"id": wfLockedID})
+	defer func() {
+		log.InfoD("pending-workflows-unlocked", logger.M{"id": wfLockedID})
+		wm.store.UnlockWorkflow(wfLockedID)
```
mohit: is there an error here we should log?

glide.yaml
rgarcia: bb39584

main.go
rgarcia: goroutines will be killed when the process gets killed, so we're good

store/dynamodb/dynamodb_store.go
```diff
@@ -495,9 +512,12 @@ func (d DynamoDB) GetWorkflowByID(id string) (resources.Workflow, error) {
 // It uses a global secondary index on the workflow name + created time for a workflow.
```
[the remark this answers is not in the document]
rgarcia: 14bb9cc

executor/workflow_manager.go
rgarcia: 👍 4898167

executor/workflow_manager.go
rgarcia: 1155c1b

rgarcia:
@mohit I think I've addressed everything

store/dynamodb/dynamodb_store.go
mohit: cc: @kofi-clever something to keep in mind as we do our planned replay testing :)

PR #44, opened 2017-09-18
support step functions

rgarcia:
## POC: workflows via AWS Step Functions

This PR creates a new implementation of the `WorkflowManager` interface that uses AWS Step Functions.

Some advantages of going the way of Step Functions:
1. It supports the entirety of the states language spec (timeouts, retries, input/output parsing, etc.)
2. We can continue to run our code in our us-west-1 ECS cluster, no network tunneling or additional log routing infra required.
3. We get a pretty good [GUI](https://us-west-2.console.aws.amazon.com/states/home?region=us-west-2#/) out of the box
4. We can swap `gearcmd` for `sfncli` and run/deploy workers the same way we do today
5. We get [CW metrics](http://docs.aws.amazon.com/step-functions/latest/dg/procedure-cw-metrics.html) on workflows, which we can alert on.

## What the PR does to the code

1. Implements a new `WorkflowManager` that uses Step Functions (`SFNWorkflowManager`).
2. Implements a new `WorkflowManager` that multiplexes across other implementations (`MultiWorkflowManager`).
3. Adds a `Manager` field to workflow definitions that is used to determine which `WorkflowManager` implementation to use.
4. Adds a `Queue` and `Namespace` field to the `Workflow` object. This was needed to identify executions in the AWS SFN API.

## Current state of this PR: I have it working locally, in theory it should work once deployed as well. How to use it locally:

1. Start this up, pointing at dev dynamo tables so we can create workflow definitions:

  ```
ark start -l -e clever-dev -s all-dev-dynamo-tables
  ```

2. Create a workflow definition with manager set to "step-functions":

  ```
curl -XPOST localhost:8080/workflow-definitions -d '{
    "name": "echo-test",
    "startAt": "echo",
    "states": [
        {
            "name": "echo",
            "resource": "echo",
            "type": "Task",
            "end": true
        }
    ],
    "manager": "step-functions"
}'
  ```

3. Start a worker locally:

  ```
sfncli -cmd echo -name raf-echo -region us-west-2
  ```

4. Begin a workflow:

  ```
curl -XPOST localhost:8080/workflows -d '{
  "workflowDefinition": { "name": "echo-test", "version": -1 },
  "data": ["hello", "world"],
  "namespace": "raf",
  "queue": "default"
}'
  ```

4. See the workflow succeed in the console:

image

🎉


mohit:
#41 also adds namespace and queue to the model and we migh need some rebasing after merging that in.

rgarcia:
@xavi- I think this is ready for review now...

swagger.yml
```diff
@@ -478,3 +488,74 @@ definitions:
         format: date-time
       type:
         type: string
+
+  # States Language Types: https://states-language.net/spec.html
```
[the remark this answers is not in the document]
rgarcia: IMO the more we use the auto-generated types in the `models` package, the easier it is to understand the impact of schema or API modifications. Also I think we should eventually move workflow definitions to contain a strict version of the states language document, instead of the sort-of version of it we have today

executor/workflow_manager_sfn.go
```diff
@@ -0,0 +1,265 @@
...
+						LastUpdated: *evt.Timestamp,
+					},
+				}
+				stateToJob[job.State] = job
+				jobs = append(jobs, job)
+			case sfn.HistoryEventTypeTaskStateExited:
+				stateExited := evt.StateExitedEventDetails
+				stateToJob[*stateExited.Name].JobDetail.StoppedAt = *evt.Timestamp
+				// TODO: match ActivitySucceeded / ActivityFailed events and set job status
+				stateToJob[*stateExited.Name].JobDetail.Status = resources.JobStatusSucceeded
```
mohit: Well this is super important to focus on. So that we can tell the difference between success and failure.

rgarcia: Yeah... I have a solution but it only works if the state machine has zero parallelism (basically assume once you enter a state all events relate to that state until you see a state exit event). Want to hear back from my AWS support ticket before committing it though

mohit:
In a different PR we Kofi is working to add support for `Job.Input, Job.Output` and `Workflow.Output`. We can them use fill in those fields using the sfn API as you mention in the comments here.

rgarcia:
@xavi- bump

swagger.yml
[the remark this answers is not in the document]
rgarcia: added job status updating bfb4a7b, but I'm pretty sure it won't work as soon as we have parallel states (I have an AWS support ticket open to confirm this)

executor/workflow_manager_sfn.go
rgarcia: bfb4a7b

main.go
```diff
@@ -58,14 +60,24 @@ func main() {
 		PrefixWorkflows:           c.DynamoPrefixWorkflows,
 	})
 	batch := batchclient.NewBatchExecutor(batch.New(awsSession(c)), c.DefaultBatchQueue, c.CustomBatchQueues)
-	wfm := executor.NewBatchWorkflowManager(batch, db)
+	wfmBatch := executor.NewBatchWorkflowManager(batch, db)
+	// TODO: config all of these
+	sfnRegion := "us-west-2"
+	sfnRole := "arn:aws:iam::589690932525:role/raf-test-step-functions"
```
xavi-: fyi, don't forget about this TODO

executor/workflow_manager.go
```diff
@@ -115,7 +115,9 @@ func (wm BatchWorkflowManager) UpdateWorkflowStatus(workflow *resources.Workflow
 
 // CreateWorkflow can be used to create a new workflow for a given WorkflowDefinition
 func (wm BatchWorkflowManager) CreateWorkflow(def resources.WorkflowDefinition, input []string, namespace string, queue string) (*resources.Workflow, error) {
```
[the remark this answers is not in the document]
rgarcia: 3d91eec

rgarcia:
@xavi- thanks, yeah, that logic could use a test. Going to keep it as a TODO for now so that we can get this out today, but will add the test sometime this week (21f6784)

Addressed TODO around config values here 4406bf1

main.go
rgarcia: 4406bf1

PR #45, opened 2017-09-20
Expose per-job inputs and outputs

kofi-clever:
This adds `input` and `output` fields to the `Job` model used in API responses.
Exposing additional fields on the internal `Job` model and using that in the conversion to WAG model.

Implementation work is WIP in an upcoming PR. Just splitting these up for ease of review.

- [x] Run "make generate"

### DIFF
--- a/handler.go
+++ b/handler.go
@@ -308,15 +308,18 @@ func apiWorkflowFromStore(workflow resources.Workflow) *models.Workflow {
 	jobs := []*models.Job{}
 	for _, job := range workflow.Jobs {
 		jobs = append(jobs, &models.Job{
-			ID:           job.ID,
-			CreatedAt:    strfmt.DateTime(job.CreatedAt),
-			StartedAt:    strfmt.DateTime(job.StartedAt),
-			StoppedAt:    strfmt.DateTime(job.StoppedAt),
-			State:        job.State,
-			Status:       string(job.Status),
-			StatusReason: job.StatusReason,
-			Container:    job.ContainerId,
-			Attempts:     job.Attempts,
+			ID:            job.ID,
+			Attempts:      job.Attempts,
+			Container:     job.ContainerId,
+			CreatedAt:     strfmt.DateTime(job.CreatedAt),
+			DependencyIds: job.DependencyIDs,
+			Input:         job.Input,
+			Output:        job.Output,
+			StartedAt:     strfmt.DateTime(job.StartedAt),
+			State:         job.State,
+			Status:        string(job.Status),
+			StatusReason:  job.StatusReason,
+			StoppedAt:     strfmt.DateTime(job.StoppedAt),
 		})
 	}
 
--- a/handler.go
+++ b/handler.go
@@ -308,18 +308,17 @@ func apiWorkflowFromStore(workflow resources.Workflow) *models.Workflow {
 	jobs := []*models.Job{}
 	for _, job := range workflow.Jobs {
 		jobs = append(jobs, &models.Job{
-			ID:            job.ID,
-			Attempts:      job.Attempts,
-			Container:     job.ContainerId,
-			CreatedAt:     strfmt.DateTime(job.CreatedAt),
-			DependencyIds: job.DependencyIDs,
-			Input:         job.Input,
-			Output:        job.Output,
-			StartedAt:     strfmt.DateTime(job.StartedAt),
-			State:         job.State,
-			Status:        string(job.Status),
-			StatusReason:  job.StatusReason,
-			StoppedAt:     strfmt.DateTime(job.StoppedAt),
+			ID:           job.ID,
+			Attempts:     job.Attempts,
+			Container:    job.ContainerId,
+			CreatedAt:    strfmt.DateTime(job.CreatedAt),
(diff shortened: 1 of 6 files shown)


resources/jobs.go
```diff
@@ -20,21 +20,24 @@ const (
...
+	StatusReason  string
+	StoppedAt     time.Time
 }
 
 // Job represents an active State and as part of a Job
 type Job struct {
 	JobDetail
 	ID            string
 	Name          string
-	Input         []string
```
kofi-clever: moved `Input` into the `JobDetail` since i'll be updating it whenever we update job status

handler.go
```diff
@@ -308,15 +308,18 @@ func apiWorkflowFromStore(workflow resources.Workflow) *models.Workflow {
...
-			State:        job.State,
-			Status:       string(job.Status),
-			StatusReason: job.StatusReason,
-			Container:    job.ContainerId,
-			Attempts:     job.Attempts,
+			ID:            job.ID,
+			Attempts:      job.Attempts,
+			Container:     job.ContainerId,
+			CreatedAt:     strfmt.DateTime(job.CreatedAt),
+			DependencyIds: job.DependencyIDs,
```
mohit: why do we need dependencyIDs in the API. for now everything is linear and is returned in the order that the tasks were executed. Also as we build more complex branching the notion of dependencyIDs will change. In summary if possible I think we should not add this to the API since it seems unique to how we handle Job flow in AWS Batch and not a Workflow Manager concept.

resources/jobs.go
```diff
@@ -20,21 +20,24 @@ const (
...
-	ContainerId  string // identification string for the running container
-	StatusReason string
-	Status       JobStatus
-	Attempts     []*models.JobAttempt
+	Attempts      []*models.JobAttempt
+	ContainerId   string // identification string for the running container
+	CreatedAt     time.Time
+	DependencyIDs []string
+	Input         []string
+	Output        string
```
mohit: I think we should keep `input` and `output` the same formats for simplicity. What do you think?

swagger.yml
```diff
@@ -388,13 +388,26 @@ definitions:
     properties:
       id:
         type: string
+      attempts:
+        type: array
+        items:
+          $ref: '#/definitions/JobAttempt'
+      Container:
```
mohit: doh this should be `container`. Well a follow up fix sometime.

handler.go
kofi-clever: yeah, that makes sense to me. meant to just expose it here and not in the actual API.
only needed to expose it as part of this model to facilitate extracting inputs from a job dependency's outputs (see https://github.com/Clever/workflow-manager/pull/46)

i can switch things around in that PR and move the input-setting logic into the `batchclient` package if we really don't want to have dependency IDs being persisted in the workflows table

resources/jobs.go
kofi-clever: https://github.com/Clever/workflow-manager/pull/45/commits/9a7f4f8ae2f8219d9ad27e4a5b3763368976622a

swagger.yml
mohit: If so let’s change it

kofi-clever:
thx! removed `DependencyIDs` for now - just going to move the logic for setting inputs into `UpdateStatus` until we know what's going to be common across Batch & Step Functions

rgarcia:
@kofi-clever @mohit why is output an array of strings? My understanding is that it will always be a JSON object https://states-language.net/spec.html#filters

mohit:
@rgarcia currently an array of strings works well for both commands that require raw args (`--arg arg`) and those with JSON objects. This is encoded in the way `ark wfbeta` is used, how `Workflow.Input` is set when creating a new Workflow, and thus how CronAdmin is setup to create a new WorkflowManager cronjob.

`sfn-cli`, `batchcli` and Jobs can use Object notation by just passing the JSON object. We just encode that as the first element of the `[]string`.   

To minimize changes in the external interface to Workflows for now I think we should stick to having all inputs and outputs as `[]string` in the interface. I'm not arguing this is the perfect decision but do think it's good for now. I believe there might be some differences in conventions around the input/output support by `sfn` and the input/output we want to support using Workflows. I don't want to change things to just support the sfn format right now. 

Happy to chat more in person.

rgarcia:
IMO it'd be truer to the states language spec to have input / output be a JSON string. It'd also limit the extent of the conventions we have around translating JSON to / from program input / output to only exist within `sfncli`/`batchcli`. E.g., `sfncli` could try to decode a top level JSON array (`[]string`) out of input and use that as args to the command, otherwise it tries to decode it as a top level JSON object (`map[string]interface{}`) and uses that as the single argument to the command.

kofi-clever:
> IMO it'd be truer to the states language spec to have input / output be a JSON string.

👍  FWIW, my vote would also be for standardization on JSON string inputs and outputs

mohit:
Just talked with Raf and he is willing to fix this end to end. i.e. changing the data model, figuring out the changes required in the API, and the conventions in the `cli`s to handle `[]string` vs. `JSON Object`

kofi-clever:
🎉

PR #46, opened 2017-09-21
Implementation for storing/propagating job inputs and outputs.

kofi-clever:
**JIRA ID:** [INFRA-2532](https://clever.atlassian.net/browse/INFRA-2532)

- When updating status (just Batch executor for now), fetch job results from the `worker-results` dynamo table.
  - (Logic mostly copied from batch-cli.)
- Store any available outputs as part of the `Job` model in the `workflows` dynamo table.
- For any jobs with dependencies, copy outputs of all dependencies to the `Job.Input` array.

Flexible on where the various pieces of code are located.

**Rollout:**
- [N/A] Run "make generate"


executor/batchclient/batch.go
```diff
@@ -11,46 +11,67 @@ import (
...
 	}
 }
 
-func (be BatchExecutor) Status(tasks []*resources.Job) []error {
-	status := map[string]*resources.Job{}
-	awsJobs := []*string{}
+func (be BatchExecutor) UpdateStatus(tasks []*resources.Job) []error {
+	tasksByID := map[string]*resources.Job{}
+	awsJobIDs := []*string{}
 	taskIds := []string{}
```
mohit: and this is `jobIDs`. Sorry but seems like we missed some spots in the big refactor

executor/batchclient/batch.go
```diff
@@ -145,16 +190,61 @@ func (be BatchExecutor) getJobDetailFromBatch(job *batch.JobDetail) (resources.J
...
+
+	fetchKeys := []map[string]*dynamodb.AttributeValue{}
+	for _, job := range jobs {
+		fetchKeys = append(fetchKeys, map[string]*dynamodb.AttributeValue{
+			"Key": {S: aws.String(job.ID)},
+		})
+	}
+
+	results, err := be.resultsDB.BatchGetItem(&dynamodb.BatchGetItemInput{
+		RequestItems: map[string]*dynamodb.KeysAndAttributes{
```
mohit: Wonder if we should just make the query against both tables for all the items and then just use the ones we got in the result. This would make it possible for us to pull data from both dev and prod tables without having to know exactly what the `namespace` for the job is set to. The more I think about it it seems like maybe we just need to make the results table a single table anyways with a `SecondaryKey` for `Namespace`. 

For now I think doing the extra work of fetching ids from both tables should work. What do you think?

executor/batchclient/batch.go
```diff
@@ -97,8 +133,9 @@ func (be BatchExecutor) Cancel(tasks []*resources.Job, reason string) []error {
 
 // getJobDetailFromBatch converts batch.JobDetail to resources.JobDetail
 func (be BatchExecutor) getJobDetailFromBatch(job *batch.JobDetail) (resources.JobDetail, error) {
-	var statusReason, containerArn string
+	var statusReason, containerArn, queueName string
 	var createdAt, startedAt, stoppedAt time.Time
+	var inputsFromEnv []string
```
mohit: where is this being set?

executor/executor.go
```diff
@@ -5,6 +5,6 @@ import "github.com/Clever/workflow-manager/resources"
 // Executor defines the interface for submitting and managing workflows
 type Executor interface {
 	SubmitWorkflow(name string, definition string, dependencies []string, input []string, queue string, attempts int64) (string, error)
-	Status(ids []*resources.Job) []error
+	UpdateStatus(ids []*resources.Job) []error
```
mohit: if this is just cosmetic we should either post in #44 that the interface is changing or just wait until that is merged.

executor/workflow_manager.go
```diff
@@ -51,11 +51,31 @@ func (wm BatchWorkflowManager) UpdateWorkflowStatus(workflow *resources.Workflow
...
+	for _, job := range workflow.Jobs {
+		if len(job.DependencyIDs) == 0 {
+			continue
+		}
+
+		job.Input = []string{}
+		for _, dependencyID := range job.DependencyIDs {
+			dependency := jobsByID[dependencyID]
+			if len(dependency.Output) > 0 {
+				job.Input = append(job.Input, dependency.Output)
```
mohit: with the new changes this should be `job.Input = append(job.Input, depdendency.Output...)` right?

executor/workflow_manager.go
```diff
@@ -51,11 +51,31 @@ func (wm BatchWorkflowManager) UpdateWorkflowStatus(workflow *resources.Workflow
 	}
 
 	// fetch new status from batch
-	errs := wm.executor.Status(workflow.Jobs)
+	errs := wm.executor.UpdateStatus(workflow.Jobs)
 	if len(errs) > 0 {
 		return fmt.Errorf("Failed to update status for %d jobs. errors: %s", len(errs), errs)
 	}
 
+	// Set job inputs based on outputs from dependencies.
```
mohit: if you think it's straightforward, let's pull this out into it's own function. My sense is that we will be revisiting this and it'll be easier if it's a independently testable function.

kofi-clever:
@mohit - rebased to the latest master to merge with the executor changes

also added logic to only set the input if the dependency succeeded, so we don't have invalid inputs showing up for aborted jobs - https://github.com/Clever/workflow-manager/pull/46/commits/1e87dffb51d5cf12b2071bddabb8146f878883df

executor/batchclient/batch.go
```diff
@@ -11,46 +11,67 @@ import (
...
+	DependenciesEnvVarName = "_BATCH_DEPENDENCIES"
 
-// StartingInputEnvVarName is used to pass in the input for the first task
-// in a job
-const StartingInputEnvVarName = "_BATCH_START"
+	// StartingInputEnvVarName is used to pass in the input for the first task
+	// in a job
+	StartingInputEnvVarName = "_BATCH_START"
+
+	productionQueueName = "workflows-production"
```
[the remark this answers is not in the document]
mohit: can we just use the `DefaultQueue` for this?

executor/batchclient/batch.go
```diff
@@ -11,46 +11,67 @@ import (
...
-	customQueues map[string]string
+	client          batchiface.BatchAPI
+	defaultQueue    string
+	customQueues    map[string]string
+	resultsDB       dynamodbiface.DynamoDBAPI
+	resultsDBConfig *ResultsDBConfig
+}
+
+type ResultsDBConfig struct {
+	TableNameDev  string
```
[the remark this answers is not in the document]
mohit: add TODO to remove this prod/dev table names.

executor/batchclient/batch.go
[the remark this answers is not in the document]
kofi-clever: ah not sure - what's the condition we'd use there? would this work?:
```go
if job.QueueName == DefaultQueue {
  // use prod table
} else {
  // use dev table
}
```

IIUC, the "default queue" setting is environment-specific. i.e. it's static for a given deployment - but in this case, we need to determine which queue the job ran in to figure out which results table to look in, right?

kofi-clever:
as discussed, will go ahead and merge this for the sake of iterating on client development.
can open a task for cleanup if there isn't one tracking the migration from batch to step fn already

PR #47, opened 2017-09-22
Add Resume Workflow By ID support

mohit:
# Implementation 

`handler.go`: `ResumeWorkflowByID`

* Only Workflows that are in a stable state (i.e. `WorkflowIsDone`) are allowed to be Resumed.

* We fetch the input from the existing `Job` for the provided `StartAt` state . If the `Job` did not start in the provided Workflow then we fail. 

`workflow_manager_{sfn,batch}`: `RetryWorkflow`

* RetryWorkflow is agnostic of the status of the original Workflow Jobs and input (that is the responsibility of the handler).

* Clone the WorkflowDefinition from the original Workflow, set the new `StartAt` state and trim the WorkflowDefinition appropriately. 

* Mimic `StartWorkflow` with the new WorkflowDefinition and the input

* Update `newWorkflow.RetryFor` and `originalWorkflow.Retires` and save both the models to the store.

## Changes to existing behaviors

1. In SFN, we now save the `Workflow` in the DB before starting an AWS Execution. This is to ensure that we don't have the case where an execution is running without workflowManager having a reference to it. 

2. StateMachineName for SFN no longer uses `queue`, since that is a meaningless distinction in Step Functions. `Queue` is replaced with `StartAt` allowing us to create StateMachines that come from the same WorkflowDefinition but have different `StartAt` states. 

3. We now rely on `Workflow` saving the full `WorkflowDefinition` (this was already being done) since changing the StartAt does not change the `WorkflowDefinition.Name` or `WorkflowDefinition.Version` but the resulting Workflow is created using a modified WorkflowDefinition. There could be a very strong argument made for storing the new WorkflowDefinition to the WorkflowDefinitions table, but I'd want to leave that thinking for the future (I believe we can move to that model without much pain) but I don't want to do a complete refactor of how we store WorkflowDefinitions as part of this change. 

# Proposed Design:

We should allow creating a "new" Workflow based on an existing Workflow when it's complete `failed|cancelled|successful`. We can do this by allowing:

```
POST /workflows/<workflow-id>

overrides:
  startAt: <new-start-at-state>
```

eventually we should also support overrides for a new Workflow:

```
POST /workflows

input: {}
overrides:
  startAt: <new-start-at-state>
```

This would create a new Workflow with a temporary WorkflowDefinition that has the new `startAt` state. We could also decide to keep the original WorkflowDefinition-id but apply the `overrides` when creating the Workflow in with an `Executor` (Batch or SFn).

## Why support this feature?

Even though external clients could use `overrides` on `POST /workflows` to get all of the functionality of "POST /workflows/<workflow-id>`, we could standardize various use cases if we allow this in workflow manager. In particular `POST /workflows/<workflow-id>` would do the following:

- use the input to the `StartAt` state in the `<workflow-id>`
- create a new Workflow and populate the field  `Workflow.RetryFor`: `<workflow-id>`
- add/append a new tag value to the workflow `Workflow.Retries`: `[]string{new-workflow-id}` 

This should allow use cases to 

- run a workflow again
- start a failed workflow from the state that failed
- cancel workflows, and later run resume their runs (aka pause and resume).
- traverse the graph of re-runs to get a sense for the different runs of related workflows.  
  
------

- [ ] Run "make generate"


mohit:
cc: @kofi-clever @rgarcia I should get this done by end of day today, any thoughts on the general plan most welcome.

swagger.yml
```diff
@@ -195,6 +195,26 @@ paths:
...
+      parameters:
+        - name: workflowId
+          in: path
+          type: string
+          required: true
+        - name: override
+          in: body
+          required: true
+          schema:
+            $ref: '#/definitions/WorkflowDefinitionOverrides'
```
rgarcia: neat... what other overrides do you imagine supporting?

swagger.yml
```diff
@@ -195,6 +195,26 @@ paths:
           type: string
         404:
           $ref: "#/responses/NotFound"
+    post:
+      summary: Start a new Workflow using job outputs of a completed Workflow from the provided position
+      operationId: resumeWorkflowByID
+      parameters:
+        - name: workflowId
```
rgarcia: nit: `workflowID`

swagger.yml
```diff
@@ -499,3 +519,9 @@ definitions:
         format: date-time
       type:
         type: string
+
+  WorkflowDefinitionOverrides:
+    type: object
+    properties:
+      startAt:
```
rgarcia: nit: `StartAt`

kofi-clever: shouldn't JSON fields be camelCased for consistency?

swagger.yml
```diff
@@ -195,6 +195,26 @@ paths:
...
           $ref: "#/responses/NotFound"
+    post:
+      summary: Start a new Workflow using job outputs of a completed Workflow from the provided position
+      operationId: resumeWorkflowByID
+      parameters:
+        - name: workflowId
+          in: path
+          type: string
+          required: true
+        - name: override
```
kofi-clever: nit: `overrides`?

handler.go
```diff
@@ -232,6 +232,76 @@ func (h Handler) CancelWorkflow(ctx context.Context, input *models.CancelWorkflo
...
+						job.ID, job.State, workflow.ID)
+			}
+
+			effectiveInput = job.Input
+			break
+		}
+	}
+
+	// TODO need `queue` and `namespace` added to the Workflow model
+	// TODO: we should allow them in the Override
```
kofi-clever: could be convinced otherwise, but don't think it fits to support "resuming" a workflow in a different namespace. feels like clients should just explicitly re-post the workflow in the other namespace if needed for some debugging purpose.

handler.go
```diff
@@ -232,6 +232,76 @@ func (h Handler) CancelWorkflow(ctx context.Context, input *models.CancelWorkflo
...
+	workflow, err := h.store.GetWorkflowByID(input.WorkflowId)
+	if err != nil {
+		return &models.Workflow{}, err
+	}
+
+	// don't allow resume if workflow is still active
+	if !workflow.IsDone() {
+		return &models.Workflow{}, fmt.Errorf("Workflow %s active: %s", workflow.ID, workflow.Status)
+	}
+	if input.Override == nil {
```
kofi-clever: nit: i don't think execution would get this far if the post body's empty, since it's marked as required

handler.go
```diff
@@ -232,6 +232,76 @@ func (h Handler) CancelWorkflow(ctx context.Context, input *models.CancelWorkflo
...
+		return &models.Workflow{}, fmt.Errorf("Must provide override")
+	}
+	if input.Override.StartAt == "" {
```
kofi-clever: can mark this as required in the swagger yaml instead to avoid custom validation

handler.go
```diff
@@ -232,6 +232,76 @@ func (h Handler) CancelWorkflow(ctx context.Context, input *models.CancelWorkflo
...
+			}
+
+			effectiveInput = job.Input
+			break
+		}
+	}
+
+	// TODO need `queue` and `namespace` added to the Workflow model
+	// TODO: we should allow them in the Override
+	newWorkflow, err := h.manager.CreateWorkflow(newDef, effectiveInput, "", "")
```
kofi-clever: what are your thoughts on preserving job execution info for jobs prior to the `startAt` point?

swagger.yml
rgarcia: `StartAt` makes it consistent with the states language `StartAt` field (states language uses  PascalCase)

handler.go
mohit: I think of that as a v2 feature for this. I want to build it in a manner that is independent of the backend but did not come up with a way to do that in this first pass. Will look into it again after I clean this up.

swagger.yml
mohit: Will make this `StartAt` in anticipation of our other fields being the same as states language.

swagger.yml
mohit: Eventually I assume we would supporting `state.resource` overrides so that workflows can target differently resources workers. But to be honest I'm not sure. Even if we don't add much else it just seemed to me better scoping to put this as an `override` rather than just adding the `StartAt` field. Happy to change if you have a strong preference.

mohit:
Apologies for the delays here. Ended up looking into #49 and code reviews. Changed the mini-spec from using `Tags` for links between retries and the original run to:

>
* create a new Workflow and populate the field Workflow.RetryFor: <workflow-id>
* add/append a new tag value to the workflow Workflow.Retries: []string{new-workflow-id}

Suggestions about the naming welcome.

swagger.yml
kofi-clever: ah gotcha. does the client-facing API need to be that closely tied with the backend implementation?
(i think i'm sold on consistency with states language - just playing devil's advocate)

also - currently, the workflow definitions have regular camel-cased fields (e.g https://github.com/Clever/ark-config/blob/master/workflows/materialize-building-blocks.json)
maybe we migrate those as well eventually to keep things consistent?

mohit: In my opinion our client side API should accept a strict subset (and one day complete) StatesLanguage wherever possible. In those terms states-language is our input mechanism and not just a backend concern. 

Correct, I'm thinking we should change the workflow definitions eventually too.

handler.go
mohit: I can not add `required:true` to the swagger.yml for this as it throws an error. Maybe I'm doing something wrong. Will check in with someone about this

kofi-clever: ah, i think it follows a different format for object properties: e.g. https://github.com/Clever/wag/blob/master/swagger/testyml/requiredfield.yml#L13-L14

IIRC though, the downside is that the field gets turned into a pointer (so that wag can check for absence vs empty)

mohit:
@kofi-clever @rgarcia I've tested the SFN implementation manually and am working on tests, but want to get the ball rolling on this. 

One thing I changed was to remove `queue` from the `stateMachineName` and add `StartAt` to it.

I'm probably going to focus on more manual testing and finding edge cases with the `billing workflow` as my highest priority tomorrow, but will get back to writing the tests here by Monday. Meanwhile if you think the code looks good or there are any changes you'd like me to make I can also get this merged in tomorrow.

executor/workflow_manager_batch.go
```diff
@@ -129,6 +129,51 @@ func (wm BatchWorkflowManager) CreateWorkflow(def models.WorkflowDefinition, inp
...
+	newDef := *ogWorkflow.WorkflowDefinition
+	newDef.StateMachine = &(*ogWorkflow.WorkflowDefinition.StateMachine)
+	newDef.StateMachine.StartAt = startAt
+
+	workflow := resources.NewWorkflow(&newDef, input, ogWorkflow.Namespace, ogWorkflow.Queue, ogWorkflow.Tags)
+	logWorkflowStatusChange(workflow, "")
+
+	stateResources, err := wm.getStateResources(workflow, ogWorkflow.Namespace)
+	if err != nil {
+		return &models.Workflow{}, err
```
kofi-clever: should we be returning `nil` here for the workflow instead?

executor/workflow_manager_batch.go
```diff
@@ -129,6 +129,51 @@ func (wm BatchWorkflowManager) CreateWorkflow(def models.WorkflowDefinition, inp
...
+
+	ogWorkflow.Retries = append(ogWorkflow.Retries, workflow.ID)
+	workflow.RetryFor = ogWorkflow.ID
+
+	// TODO: fails we should either
+	// 1. reconcile somehow with the scheduled jobs
+	// 2. kill the running jobs so that we don't have orphan jobs in AWS Batch
+	err = wm.store.SaveWorkflow(*workflow)
+	if err != nil {
+		return workflow, err
```
kofi-clever: `s/workflow/(nil|&models.Workflow{})/`?

executor/workflow_manager_batch.go
```diff
@@ -163,13 +208,17 @@ func (wm BatchWorkflowManager) CancelWorkflow(workflow *models.Workflow, reason
 func (wm BatchWorkflowManager) scheduleJobs(workflow *models.Workflow,
 	stateResources map[string]*models.StateResource, input string, queue string) error {
 
-	jobs := map[string]*models.Job{}
+	if err := resources.RemoveInactiveStates(workflow.WorkflowDefinition.StateMachine); err != nil {
```
kofi-clever: not related to this PR, but since this really only applies to dynamically created definitions, should we also have validation logic in SaveWorkflowDefinition to prevent people from creating definitions with orphaned states?

executor/workflow_manager_batch_test.go
```diff
@@ -266,6 +264,51 @@ func TestCreateWorkflow(t *testing.T) {
...
+	err = store.SaveWorkflow(*workflow)
+
+	t.Log("Retry an existing Workflow if it is done")
+	newWorkflow, err := jm.RetryWorkflow(*workflow, "second-state", `["args"]`)
+	assert.Nil(t, err)
+	ogWorkflow, err := store.GetWorkflowByID(workflow.ID)
+
+	assert.Equal(t, workflow.ID, ogWorkflow.ID)
+	fmt.Println(ogWorkflow)
+	fmt.Println(newWorkflow)
```
kofi-clever: nit: reminder to clean up debug logging

executor/workflow_manager_sfn.go
```diff
@@ -69,21 +69,21 @@ func stateMachineWithFullActivityARNs(stateMachine *models.SLStateMachine, regio
 // https://docs.aws.amazon.com/step-functions/latest/apireference/API_CreateStateMachine.html#StepFunctions-CreateStateMachine-request-name
 var stateMachineNameBadChars = []byte{' ', '<', '>', '{', '}', '[', ']', '?', '*', '"', '#', '%', '\\', '^', '|', '~', '`', '$', '&', ',', ';', ':', '/'}
 
-func stateMachineName(wdName string, wdVersion int64, namespace string, queue string) string {
-	name := fmt.Sprintf("%s--%s--%d--%s", namespace, wdName, wdVersion, queue)
+func stateMachineName(wdName string, wdVersion int64, namespace string, startAt string) string {
+	name := fmt.Sprintf("%s--%s--%d--%s", namespace, wdName, wdVersion, startAt)
```
kofi-clever: q: are we trying to avoid grouping retries together with full runs? wondering what the motivation is for that?

executor/workflow_manager_sfn.go
```diff
@@ -151,17 +143,73 @@ func (wm *SFNWorkflowManager) CreateWorkflow(wd models.WorkflowDefinition, input
...
+
+	err = wm.store.SaveWorkflow(*workflow)
+	if err != nil {
+		return workflow, err
+	}
+
+	// also update ogWorkflow
+	err = wm.store.UpdateWorkflow(ogWorkflow)
+	if err != nil {
+		// TODO: should we be failing here or just log?
```
kofi-clever: i would say logging would be better. at least, in the multiverse case, if the user sees an error and replays a 2nd time, it could be pretty back in the lock phase (if there are essentially 2 branches running in parallel) - we've had a flare arise from that before.

that being said, i now realise we have the same issue on L200 above where we return an error if we can't save the workflow. i think it might be save to:
- save workflow
- update parent
- start execution

if a client sees a workflow record with no execution, that's fine - we can fail automatically in WM or cancel in hubble and replay again.
if a phantom execution exists that clients don't know about, could run into issues

executor/workflow_manager_batch.go
mohit: https://github.com/Clever/workflow-manager/pull/47/commits/a777759baaf059acaa447dda63177c6d333c1957

executor/workflow_manager_sfn.go
```diff
@@ -151,17 +143,73 @@ func (wm *SFNWorkflowManager) CreateWorkflow(wd models.WorkflowDefinition, input
...
 }
 
+func (wm *SFNWorkflowManager) RetryWorkflow(ogWorkflow models.Workflow, startAt, input string) (*models.Workflow, error) {
+	if !resources.WorkflowIsDone(&ogWorkflow) {
+		return nil, fmt.Errorf("Retry not allowed. Workflow state is %s", ogWorkflow.Status)
+	}
+
+	// modify the StateMachine with the custom StartState by making a new WorkflowDefinition (no pointer copy)
+	newDef := *ogWorkflow.WorkflowDefinition
+	newDef.StateMachine = &(*ogWorkflow.WorkflowDefinition.StateMachine)
```
kofi-clever: btw - this might still be holding a reference to the ogWorkflow's state machine. i tried replaying a workflow and noticed its definition gets truncated to the point of the replay (e.g. https://mp-sfn--hubble.int.clever.com/multiverse:multiverse-workflow-definition/workflows/w/f2359d1d-d363-4fe7-904a-1ac212504a8f/jobs)

mohit: :facepalm: that was me trying to be more concise. Will revert back to deepcopy

executor/workflow_manager_sfn.go
mohit: StateMachines are immutable in step functions so to change the start state we need to create a new StateMachine. This ensures we don’t have collisions when retries

executor/workflow_manager_batch.go
mohit: https://github.com/Clever/workflow-manager/pull/47/commits/8831f4b2291c844e10e80fdb95d41fbc67271495

executor/workflow_manager_sfn.go
mohit: https://github.com/Clever/workflow-manager/pull/47/commits/77a2f37c7f6c6f0d8625ea8ab1c178f39141f3a9

executor/workflow_manager_sfn.go
mohit: @kofi-clever at least for sfn I really like the idea of saving the workflow before starting the execution, at least until we are confident about our dynamo limit configuration. We should probably add some UI detail around issues where executions fail to clean up the pending Workflows (will need to re-create that ticket about hanging workflows I just closed :))

what do you think of https://github.com/Clever/workflow-manager/pull/47/commits/91e668a1118acb8523f2a4eb4eadee83c745fce1

mohit:
@kofi-clever Think I've addressed most of the issues here, let me know what you think?

executor/workflow_manager_sfn.go
kofi-clever: ahh that makes sense.
this breaks any existing workflows though, right? (since the exection arn is changing too.) noticed some failed workflows recently that i could no longer cancel with this version of WM because the arn had changed.
should we be persisting the arn with each workflow on our end to avoid that? also fine if the answer is that the arn generation logic won't really change after this. the fallout is minimal right now, since we're mostly testing

executor/workflow_manager_batch_test.go
```diff
@@ -266,6 +264,53 @@ func TestCreateWorkflow(t *testing.T) {
...
+
+	t.Log("Retry for a running Workflow is not allowed")
+	_, err = jm.RetryWorkflow(*workflow, "second-state", `["args"]`)
+	assert.NotNil(t, err)
+
+	workflow.Status = models.WorkflowStatusFailed
+	for _, job := range workflow.Jobs {
+		job.Status = models.JobStatusFailed
+	}
+	err = store.SaveWorkflow(*workflow)
```
kofi-clever: nit: `(assert|require).NoError` for this err

executor/workflow_manager_sfn.go
mohit: hmm my expectation for right now was that we are just running tests so it's okay. Though I have been thinking about using just the `workflowID` in the `executionARN` and not the stateMachineName. That way we would not have this problem in the future. @rgarcia what do you think about making `executionARN` independent of the `stateMachineName`. That way changes in stateMachineNames (and there is a possibility to have to change that name in the future) would not affect running Executions.

mohit: Oh Sorry, looked through the code again and we already set the `ExecutionName` to `workflowID`. AWS sets the ARN for an execution to contain the `stateMachineName` so my plan is not really possible. 

@kofi-clever @rgarcia let's chat on Monday about this a little bit just to make sure we all are on the same page. I'm okay with storing the executionARN in the Workflow, but don't think we should expose that in the client and since the models are the same I'm not sure what's the best way to make that happen. 

I'm leaning towards punting this decision for now and making sure that we work on a migration plan if we end up needing to change the `stateMachineName` after our launch.

executor/workflow_manager_batch_test.go
mohit: https://github.com/Clever/workflow-manager/pull/47/commits/9b515d7dc21abe572b189eb2f51f5f20bd5106f1

mohit:
Merging, will follow up with more tests in a separate PR, and any other fixes that might come up.

PR #48, opened 2017-09-25
Break up getWorkflows into getActiveWorkflows and getInactiveWorkflows.

kofi-clever:
No implementation yet. Just API definition updates.

- Both `getActiveWorkflows` and `getInactiveWorkflows` share a common `WorkflowQuery` payload.
- For now, `WorkflowQuery` just contains a list of workflow statuses to match against and a `limit` for the page size.
- Will follow up later to add date range filtering (and default to a reasonable window, e.g. last 7 days)

- [x] Run "make generate"


swagger.yml
```diff
@@ -126,6 +126,62 @@ paths:
         404:
           $ref: "#/responses/NotFound"
 
+  /workflow-definitions/{definitionName}/workflows/active:
```
mohit: Maybe I'm being too pedantic but I would think we should have this still be `/workflow-definitions/{definitionName}` and have the `active/inactive` be a query param. example `/workflow-definitions/{definitionName}?active=true`. If it's not set then we should just return all workflows by createdTime and of course respect it when it is set.

kofi-clever: ah, was trying to avoid overly flexible queries that would require scanning over the entire index until we have a better search/filter backend in place.

could also move the active/inactive switch to the `WorkflowQuery` object to reduce the API surface area a bit, but thought the explicitness could be useful.

no strong feelings on whether or not active/inactive is optional - FWIW, it does require a bit more code complexity to support querying and filtering by both the <definitionName> and <definitionName>:<ACTIVE|INACTIVE> indexes via one API. leaning toward waiting to see user need for that before the additional investment

mohit: 1. I just think filters as `query params`  are more RESTy than implemented as `slugs`. The actual query would be the same just the swagger definition would be different. 

2. the use case I'm thinking for both `active/inactive` would be for `ark` where we probably want to show `COMPLETED` and `RUNNING` jobs together.

kofi-clever:
ok, updated the API to consolidate the 'get workflow' queries into the existing one.
should be backward-compatible, based on the generated code.

writing tests for the implementation and will send out the PR for that soon

PR #49, opened 2017-09-26
[INFRA-2535] tags support for workflows

bgveenstra:
**JIRA**: [INFRA-2535](https://clever.atlassian.net/browse/INFRA-2535)

**Overview**:

This PR:
- adds `tags` to the `Workflow` model, so that any non-nil values for these will be included in the `POST /workflows` (`StartWorkflow`) response as well as any other response that sends back a `Workflow`.
- converts the type of the tags from `map[string]interface{}` to `map[string]string` in the handler
- updates the readme after recent naming changes
- updates additional small items after naming changes

**Known Bug**:

A workflow's tags (and maybe namespace and queue - mostly saw this yesterday) are lost at some unknown point after initially being saved in ddb.  To trigger this issue, I:

- run workflow-manager locally
- start a workflow through API using `curl`
- see debugging log messages that show the workflow retrieved from dynamodb with tags intact
- wait up to 30 seconds, monitoring debugging log messages (mostly from `GetWorkflowByID`)
- at some point, the tags will become `<nil>` in the log

**Testing**:

- [x] rewrite existing tests based on API changes
- [x] ⚠️  tags (and possibly namespace and queue) are removed at some point during polling

**Rollout**:
- [x] Run "make generate"


README.md
```diff
@@ -1,6 +1,6 @@
 # workflow-manager
 
-Minimal Workflow orchestrator for AWS Batch
+Minimal Workflow orchestrator for AWS Batch and Step Functions
```
mohit: @bgveenstra thanks so much for the cleanup here!

mohit:
whoops - see that you already added the process. Thanks!

mohit:
@bgveenstra the issue here is not with your code at all, but the fact that the default dev DBs are shared with the workflow-manager running in `clever-dev`. The polling there was removing `Tags`

PR #50, opened 2017-09-26
Filter/pagination implementation for getWorkflows API

kofi-clever:
Adds support for sorting/filtering/pagination to the `getWorkflows` API.
Uses a new secondary index on the concatenated workflow definition name and status: https://github.com/Clever/infra/pull/641

- [x] Run "make generate"

### DIFF
--- a/store/dynamodb/dynamodb_store.go
+++ b/store/dynamodb/dynamodb_store.go
@@ -137,6 +137,7 @@ func (d DynamoDB) InitTables() error {
 	for _, ads := range [][]*dynamodb.AttributeDefinition{
 		(ddbWorkflowPrimaryKey{}.AttributeDefinitions()),
 		(ddbWorkflowSecondaryKeyWorkflowDefinitionCreatedAt{}.AttributeDefinitions()),
+		(ddbWorkflowSecondaryKeyDefinitionStatusCreatedAt{}.AttributeDefinitions()),
 		(ddbWorkflowSecondaryKeyStatusLastUpdated{}.AttributeDefinitions()),
 	} {
 		workflowAttributeDefinitions = append(workflowAttributeDefinitions, ads...)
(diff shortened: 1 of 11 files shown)


store/dynamodb/dynamodb_store.go
```diff
@@ -540,37 +552,61 @@ func (d DynamoDB) GetWorkflowByID(id string) (resources.Workflow, error) {
...
+		return workflows, nextPageToken, err
 	}
 
 	for _, item := range res.Items {
 		workflow, err := DecodeWorkflow(item)
 		if err != nil {
-			return workflows, err
+			return workflows, nextPageToken, err
 		}
 		if err := d.populateWorkflow(&workflow); err != nil {
```
mohit: not your change but this should be `populateWorkflowDefinition(&workflow)` also we should add a TODO or comment that we could use `ddb.BatchGet` for this eventually.

store/dynamodb/page_key.go
```diff
@@ -0,0 +1,63 @@
+package dynamodb
```
mohit: FYI: I'm not familiar with paging much but looks good from my perspective

store/store.go
```diff
@@ -8,7 +8,7 @@ import (
 	"github.com/Clever/workflow-manager/resources"
 )
 
-// WorkflowStore defines the interface for persistence of Workflow defintions
+// Store defines the interface for persistence of Workflow definitions.
```
mohit: s/Workflow definitions/Workflow Manager resources/

store/tests/tests.go
```diff
@@ -199,29 +200,113 @@ func GetWorkflowByID(s store.Store, t *testing.T) func(t *testing.T) {
...
+	}
+}
 
-		workflows, err := s.GetWorkflows(wf2.Name())
-		require.Nil(t, err)
-		require.Equal(t, len(workflows), len(workflow2IDs))
-		var gotWorkflow2IDs []string
-		for _, j := range workflows {
-			gotWorkflow2IDs = append(gotWorkflow2IDs, j.ID)
+func GetWorkflowsPagination(s store.Store, t *testing.T) func(t *testing.T) {
```
mohit: should we test for some pagination error cases here?

store/dynamodb/dynamodb_store.go
kofi-clever: agreed - https://github.com/Clever/workflow-manager/pull/50/commits/e840927ec091bd7e5ed5b581c7d994345812a1bf

store/store.go
kofi-clever: oh yeah, good catch - https://github.com/Clever/workflow-manager/pull/50/commits/907f371759bd1c7450ebd587f987ae9ef7516bc9

store/tests/tests.go
kofi-clever: +1 - https://github.com/Clever/workflow-manager/pull/50/commits/d22346cf0ea6da0a2fc1b0977f13c695d95c306a

PR #51, opened 2017-09-26
tweak naming of state machines / resources

rgarcia:
Use double dash `--` to separate the different components of state machine names and resource names


bgveenstra:
LGTM

PR #52, opened 2017-09-28
schema changes

rgarcia:
WorkflowDefinition:
  - stateMachine: new top level field that is an SLStateMachine (the subset of states language that we support)
  - startAt -> stateMachine.StartAt
  - description -> stateMachine.Comment
  - states (array) -> stateMachine.States (object)

Job:
 - input (array of string) -> input (string)
 - output (array of string) -> output (string)

NewWorkflowDefinitionRequest:
  - stateMachine: new top level field that is an SLStateMachine
  - startAt -> stateMachine.StartAt
  - description -> stateMachine.Comment
  - states (array) -> stateMachine.States (object)

ErrorEquals -> SLErrorEquals (to make clear that this is a states language construct)

workflowId -> workflowID

State deleted, since we use SLState
Retrier deleted, since we use SLRetrier

SLState:
- add TimeoutSeconds and HeartbeatSeconds in preparation for supporting these


rgarcia:
cc @mohit @kofi-clever @xavi- opening this for review before I proceed further down the path

swagger.yml
```diff
@@ -317,55 +317,10 @@ definitions:
     properties:
```
mohit: needs `state-machine`

swagger.yml
```diff
@@ -601,3 +549,9 @@ definitions:
       - "Succeed"
       - "Fail"
       - "Parallel"
+
+  SLErrorEquals:
+    type: string
+    enum:
```
mohit: I do think we will be adding our own ErrorEqual enums here, just FYI/for discussion.

swagger.yml
```diff
@@ -439,13 +388,9 @@ definitions:
         type: string
         format: date-time
       input:
-        type: array
-        items:
-          type: string
+        type: string
```
mohit: add comment that it's a JSON marshaled as string

mohit:
This seems like a good step to me. maybe we can even support the `state-machine` to be yaml eventually wrapped in the JSON of `NewWorkflowRequest` :p

swagger.yml
rgarcia: that's consistent with states language spec

```
Errors are identified by case-sensitive strings, called Error Names. The States language defines a set of built-in strings naming well-known errors, all of which begin with the prefix “States.”
```

so really we should probably change this to be a string and remove the enum

swagger.yml
rgarcia: 8f45465

swagger.yml
rgarcia: 6f06f3c

rgarcia:
@mohit re-generated the code and have the tests passing. Ended up gutting the types out of the `resources` package in favor of `models` types, since (1) mirroring the schema changes in the `resources` types was more work than just removing them, and (2) if we're going to be changing how data looks in dynamo, we might as well move off of the gob encoding that the types in the `resources` package required.

My plan from here is to set up `v3` tables in dynamo, write a script that migrates data over, and then start some manual testing.

mohit:
@rgarcia going to merge #49 so that we can get those changes in too. don't think it should cause much issues merging with your changes.

rgarcia:
@mohit (painfully) rebased

mohit:
Whoops. Apologies I’ll rebase the redo branch on your changes after it’s done.

PR #53, opened 2017-10-01
Add getWorkflows query options for leaning out the response

kofi-clever:
1. Remove the `UpdateStatus` step from the `getWorkflows` endpoint. `getWorkflowByID` will continue to return the most up-to-date status.
2. Add `?includeFullDefinition` and `?includeJobs` (both `false` by default) for optionally skipping the DB query for full workflow definition data and excluding job data respectively for clients that just need a summary of the workflows.
  Helps keep queries for large sets of workflows performant and keep the memory footprint low for clients paging through results.

- [x] Run "make generate"


gen-go/models/inputs.go
```diff
@@ -305,6 +307,14 @@ func (i GetWorkflowsInput) Path() (string, error) {
 	path := "/workflows"
 	urlVals := url.Values{}
 
+	if i.IncludeFullDefinition != nil {
+		urlVals.Add("includeFullDefinition", strconv.FormatBool(*i.IncludeFullDefinition))
+	}
+
+	if i.IncludeJobs != nil {
```
mohit: My sense is that this does not add any optimizations since it's in the same table.

store/dynamodb/dynamodb_store.go
```diff
@@ -596,12 +596,17 @@ func (d DynamoDB) GetWorkflows(query *store.WorkflowQuery) ([]resources.Workflow
...
 			return workflows, nextPageToken, err
 		}
-		// TODO: Optimization - do a BatchGet to fetch the workflow definitions all at once and avoid
-		// redundant fetches.
-		// TODO: Add an option in the API to skip this step for clients that want to fetch definitions
-		// separately.
-		if err := d.attachWorkflowDefinition(&workflow); err != nil {
-			return workflows, nextPageToken, err
+
+		if !query.IncludeJobs {
```
mohit: I don't see a benefit of clearing out data we have already fetched from the DB?

rgarcia:
if possible I'd like to wait until #52 is merged before changing master... I've gone through one rebase on #52 and it was not easy. ETA on merging #52 is tomorrow

store/dynamodb/dynamodb_store.go
kofi-clever: ah, this is more of a memory optimisation for clients - my reasoning:
- job data can fairly bulky, especially given the inputs and outputs
- using multiverse as a sample case, the payload for 50k+ workflows during a flare is easily upwards of 0.5GB in memory (vs ~30MB without job and definition data). and that's a single user opening up hubble, so that gets unnecessarily out of hand with multiple users logged in.

could just scale up hubble when needed, but figured if it's this simple to avoid, might as well. wdyt?

swagger.yml
```diff
@@ -151,6 +151,16 @@ paths:
         pageParameter: pageToken
       parameters:
         # TODO: Add date range filters.
+        - name: includeFullDefinition
+          description: Whether or not ot include the full workflow definition
```
mohit: s/ot//g

store/dynamodb/dynamodb_store.go
mohit: Thanks so much for describing the issue kofi! 

It seems like you're more looking for a summary view so wonder if we should add a single `summary=true` that turns off both workflow definitions and job details.

kofi-clever: yeah, i like that! i'm actually using separate `WorkflowSummary` and `DetailedWorkflow` objects in hubble, so this would fit nicely with those semantics - https://github.com/Clever/workflow-manager/pull/53/commits/e955fd1857172140f00dafa45389e55a7c84e0ec

swagger.yml
kofi-clever: thx!

store/dynamodb/dynamodb_store.go
```diff
@@ -596,12 +596,17 @@ func (d DynamoDB) GetWorkflows(query *store.WorkflowQuery) ([]resources.Workflow
...
 			return workflows, nextPageToken, err
 		}
-		// TODO: Optimization - do a BatchGet to fetch the workflow definitions all at once and avoid
-		// redundant fetches.
-		// TODO: Add an option in the API to skip this step for clients that want to fetch definitions
-		// separately.
-		if err := d.attachWorkflowDefinition(&workflow); err != nil {
-			return workflows, nextPageToken, err
+
+		if query.SummaryOnly {
```
mohit: TODO: don't pull job data from dynamodb if it's not required.

kofi-clever: 💯

mohit:
Tests are failing but generally this looks good to go once tests are fixed

rgarcia:
not sure if you noticed in rebase, but new schema stores workflowdefinition as an embedded object in the workflow, so the `attach...` method is gone, which might make the inefficiency problems less of an issue

kofi-clever:
> not sure if you noticed in rebase, but new schema stores workflowdefinition as an embedded object in the workflow

yeah! made this change a little cleaner than before. and the latency doesn't degrade anymore when ?summaryOnly=false which is great

PR #54, opened 2017-10-02
Workflow-id is now added to the jobs before they're submitted

xavi-:
Workflow-id is now added as property as `$execution-name`.  It's used by sfncli and kayvee to make it easier to correlate logs from a workflow across many workers.

Also added validation to ensure that the worker input is a json object.

Related: https://github.com/Clever/sfncli/pull/4
Jira: https://clever.atlassian.net/browse/INFRA-2553, https://clever.atlassian.net/browse/INFRA-2526

### DIFF
--- a/executor/workflow_manager_sfn.go
+++ b/executor/workflow_manager_sfn.go
@@ -130,7 +130,7 @@ func (wm *SFNWorkflowManager) CreateWorkflow(wd models.WorkflowDefinition, input
 	if err := json.Unmarshal([]byte(input), &inputJSON); err != nil {
 		return nil, fmt.Errorf("input is not a valid JSON object: %s", err)
 	}
-	inputJSON["_EXECUTION_NAME"] = &executionName
+	inputJSON["_EXECUTION_NAME"] = *executionName
 
 	marshaledInput, err := json.Marshal(inputJSON)
 	if err != nil {
--- a/executor/workflow_manager_sfn.go
+++ b/executor/workflow_manager_sfn.go
@@ -125,25 +125,24 @@ func (wm *SFNWorkflowManager) CreateWorkflow(wd models.WorkflowDefinition, input
 	workflow := resources.NewWorkflow(&wd, input, namespace, queue, tags)
 
 	executionName := aws.String(workflow.ID)
+
+	var inputJSON map[string]interface{}
+	if err := json.Unmarshal([]byte(input), &inputJSON); err != nil {
+		return nil, fmt.Errorf("input is not a valid JSON object: %s", err)
+	}
+	inputJSON["_EXECUTION_NAME"] = &executionName
+
+	marshaledInput, err := json.Marshal(inputJSON)
+	if err != nil {
+		return nil, err
+	}
+
+	// We ensure the input is a json object, but for context the
 	// observed StartExecution behavior for different edge-case inputs:
 	// - nil: AWS converts this to an input of an empty object "{}"
 	// - aws.String(""): leads to InvalidExecutionInput AWS error
 	// - aws.String("[]"): leads to an input of an empty array "[]"
-	var startExecutionInput *string
-	if len(input) > 0 {
-		var inputJSON map[string]interface{}
-		if err := json.Unmarshal([]byte(input), &inputJSON); err != nil {
-			return nil, fmt.Errorf("input is not a valid JSON object: %s", err)
-		}
-		inputJSON["_EXECUTION_NAME"] = &executionName
-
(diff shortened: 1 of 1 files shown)


executor/workflow_manager_sfn.go
```diff
@@ -129,11 +129,18 @@ func (wm *SFNWorkflowManager) CreateWorkflow(wd models.WorkflowDefinition, input
...
 	var startExecutionInput *string
 	if len(input) > 0 {
-		var inputJSON interface{}
+		var inputJSON map[string]interface{}
 		if err := json.Unmarshal([]byte(input), &inputJSON); err != nil {
-			return nil, fmt.Errorf("input is not valid JSON: %s", err)
+			return nil, fmt.Errorf("input is not a valid JSON object: %s", err)
 		}
-		startExecutionInput = aws.String(input)
+		inputJSON["$execution-name"] = workflow.ID
```
rgarcia: is `$` a convention you've seen used elsewhere? I've typically seen `_` to denote hidden fields.

rgarcia: also, since we also reference workflow ID as execution name below (`Name:            aws.String(workflow.ID),`) it might be a good idea to create a variable like `var startExecutionName *string` so that there's no risk of divergence

xavi-: In javascript parameters with that start with `$` usually meant machine-generated.  Also, I had noticed that a couple of workers already use `_`'ed fields.  I felt using a `$` made it clear that it was foreign.  I can change it to `_`.

xavi-: Renamed in 80ea378 and variable added in 
dc7852d

executor/workflow_manager_sfn.go
```diff
@@ -123,22 +123,31 @@ func (wm *SFNWorkflowManager) CreateWorkflow(wd models.WorkflowDefinition, input
...
 	// - aws.String("[]"): leads to an input of an empty array "[]"
 	var startExecutionInput *string
 	if len(input) > 0 {
-		var inputJSON interface{}
+		var inputJSON map[string]interface{}
 		if err := json.Unmarshal([]byte(input), &inputJSON); err != nil {
-			return nil, fmt.Errorf("input is not valid JSON: %s", err)
+			return nil, fmt.Errorf("input is not a valid JSON object: %s", err)
+		}
+		inputJSON["_EXECUTION_NAME"] = &executionName
```
rgarcia: I think you want `*executionName` here (`executionName` is `*string`). Although anything probably works since `json` marshals a JSON string out of `string` `*string` `**string` etc.

executor/workflow_manager_sfn.go
```diff
@@ -123,22 +123,31 @@ func (wm *SFNWorkflowManager) CreateWorkflow(wd models.WorkflowDefinition, input
...
+
+		marshaledInput, err := json.Marshal(inputJSON)
+		if err != nil {
+			return nil, err
 		}
-		startExecutionInput = aws.String(input)
+
+		startExecutionInput = aws.String(string(marshaledInput))
```
rgarcia: just read your JIRA comment and realized this logic should probably be revamped... the `if` statement was protecting against the case where `input == ""`.  Passing AWS this as input causes an error. We should probably do something like 


```golang
	inputJSON := map[string]interface{}{}
	if len(input) > 0 {
		if err := json.Unmarshal([]byte(input), &inputJSON); err != nil {
			return nil, fmt.Errorf("input is not valid JSON: %s", err)
		}
	}
	inputJSON["_EXECUTION_NAME"] = *executionName
	marshaledInput, err := json.Marshal(inputJSON)
	if err != nil {
		return nil, err
	}
	startExecutionInput := aws.String(string(marshaledInput))
```

Although, if we want to hold a hard line on input being JSON, we might want to consider removing the if statement, in which case empty string input would throw the "input is not valid JSON..." error. I'm kind of tempted to do that...

executor/workflow_manager_sfn.go
xavi-: Ahh, thanks.  Fixed in 6ba440aa0b0d7bf1809345ae68cb1b5108e42458

executor/workflow_manager_sfn.go
xavi-: I opted for making `""` and `nil` errors in 3e00ea6e75e9cd23281d67b426015e4ee5423865

rgarcia: 👍 input and output must be JSON feels right to me as well (and follows state language spec)

PR #55, opened 2017-10-03
use req.Manager in create/update WorkflowDefinition

mohit:
No external API changes. Seems like this just slipped past the mega-schema migration.

- [ ] Run "make generate"


mohit:
cc: @rgarcia

rgarcia:
👍

PR #56, opened 2017-10-03
Fix job input propagation for sfn manager.

kofi-clever:
Since the API exposes the input as a JSON string, skip the unmarshalling step and pass the input through directly.

Was previously generating empty strings.

- [N/A] Run "make generate"


rgarcia:
Ah good catch, lgtm

PR #57, opened 2017-10-03
Adjust resource limits based on patterns while testing in mp-sfn.

kofi-clever:
Seems to be struggling under current (fairly small) load.
Bump up mem and cpu and add autoscaling. Not tied to these numbers - open to change them.

- [N/A] Run "make generate"


kofi-clever:
verbal LGTM

PR #59, opened 2017-10-04
Handle ExecutionAborted and mark jobs as aborted_by_user accordingly.

kofi-clever:
When an `ExecutionAborted` event is encountered, mark the job as `aborted_by_user` (i.e. cancelled).

e.g. https://us-west-2.console.aws.amazon.com/states/home?region=us-west-2#/executions/details/arn:aws:states:us-west-2:589690932525:execution:mp-sfn--multiverse-multiverse-workflow-definition--1--development:3ee31208-f05a-45da-a537-78f403d66359

- [N/A] Run "make generate"

### DIFF
--- a/executor/workflow_manager_sfn.go
+++ b/executor/workflow_manager_sfn.go
@@ -264,6 +264,10 @@ func (wm *SFNWorkflowManager) UpdateWorkflowStatus(workflow *models.Workflow) er
 				if currentState != nil {
 					stateToJob[*currentState].Status = models.JobStatusSucceeded
 				}
+			case sfn.HistoryEventTypeExecutionAborted:
+				if currentState != nil {
+					stateToJob[*currentState].Status = models.JobStatusAbortedByUser
+				}
 			case sfn.HistoryEventTypeTaskStateExited:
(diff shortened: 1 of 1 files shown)


executor/workflow_manager_sfn.go
```diff
@@ -264,6 +264,10 @@ func (wm *SFNWorkflowManager) UpdateWorkflowStatus(workflow *models.Workflow) er
 				if currentState != nil {
 					stateToJob[*currentState].Status = models.JobStatusSucceeded
 				}
+			case sfn.HistoryEventTypeExecutionAborted:
+				if currentState != nil {
+					stateToJob[*currentState].Status = models.JobStatusAbortedByUser
```
mohit: I think this transition also has the reason in the struct

mohit: It’s called ‘cause’

kofi-clever: oh nice! will extract that in my cancellation PR.
looks like we'll have to trim it a bit because sfn basically just takes all the logs from the activity and puts it in the cause.
we can do something similar to gearman and take the last 3 log lines if there are more than 3

PR #60, opened 2017-10-05
Persist status reason when cancelling a workflow

kofi-clever:
When cancelling a workflow:
- Send StopExecution request to SFN.
- Set the `statusReason` field
- Check for the execution status and if the execution is in the `failed` state, update workflow status to `cancelled`.
  - Handles the case where the user cancels a failed workflow (to essentially mark it as resolved).

When updating workflow status:
- Noop if the workflow and all its jobs are in a final state.
  - Avoids resetting a user-cancelled workflow back to `failed`, for example.
- Update the `stoppedAt` timestamp if a job is moved into the failed or aborted state. This was previously missing.
- Copy over the ActivityFailed event cause to the workflow `statusReason`. Looks like SFN just dumps all of the worker's log output, so I'm trimming this down to 3 log lines for now. Arbitrary number, we can change as we see fit.
- For ExecutionAborted events, mark the last job as aborted if it's not in a final state.
  - Also copy the aborted cause to `Job.StatusReason` if there's one (this is currently a noop - SFN doesn't seem to persist the cancel reason). We still have it persisted on `Workflow.StatusReason` anyway.

**Sample end-states:**
Screenshot from Gyazo

Screenshot from Gyazo

- [N/A] Run "make generate"


executor/workflow_manager_sfn_test.go
```diff
@@ -44,3 +65,365 @@ func TestStateMachineName(t *testing.T) {
...
+	)
+}
+
+func (c *sfnManagerTestController) saveWorkflow(t *testing.T, workflow *models.Workflow) {
+	require.NoError(t, c.store.SaveWorkflow(*workflow))
+}
+
+func (c *sfnManagerTestController) tearDown() {
+	c.mockController.Finish()
+}
```
rgarcia: :100: tests!

rgarcia:
Re: `Looks like SFN just dumps all of the worker's log output` -> this is the current behavior of `sfncli`--it passes stderr as the `cause`. It's limited in how much it can report to 32k (see http://docs.aws.amazon.com/step-functions/latest/apireference/API_SendTaskFailure.html#API_SendTaskFailure_RequestSyntax)

I still think it makes sense to limit it here even further, but just wanted to share this context.

executor/workflow_manager_sfn.go
```diff
@@ -267,14 +310,30 @@ func (wm *SFNWorkflowManager) UpdateWorkflowStatus(workflow *models.Workflow) er
 			case sfn.HistoryEventTypeActivityFailed:
 				if currentState != nil {
 					stateToJob[*currentState].Status = models.JobStatusFailed
+					stateToJob[*currentState].StoppedAt = strfmt.DateTime(*evt.Timestamp)
+					if details := evt.ActivityFailedEventDetails; details != nil {
+						reasonLines := strings.Split(strings.TrimSpace(aws.StringValue(details.Cause)), "\n")
```
mohit: ahem. I'd be in favor of not having lines from the worker as the StatusReason at all, or maybe just the last line. Also seems like we should probably not be setting all of the output from sfncli but that's a separate discussion (as Raf pointed out).

mohit: My current reasoning: StatusReason should be short so that we can display it in context in various places such as ark and hubble and having multiple lines should just be solved by looking at logs. Can be convinced otherwise but ideally I'd like us to have something like a 50character limit here.

executor/workflow_manager_sfn.go
```diff
@@ -267,14 +310,30 @@ func (wm *SFNWorkflowManager) UpdateWorkflowStatus(workflow *models.Workflow) er
...
+					}
 				}
 			case sfn.HistoryEventTypeActivitySucceeded:
 				if currentState != nil {
 					stateToJob[*currentState].Status = models.JobStatusSucceeded
 				}
 			case sfn.HistoryEventTypeExecutionAborted:
 				if currentState != nil {
-					stateToJob[*currentState].Status = models.JobStatusAbortedByUser
+					job := stateToJob[*currentState]
```
mohit: Just for my understanding, what happens if `resources.JobIsDone() == true` and under what conditions do we expect that to be the case.

mocks/mock_sfniface/mock_sfnapi.go
```diff
@@ -0,0 +1,778 @@
+// Automatically generated by MockGen. DO NOT EDIT!
```
mohit: is this supposed to be committed or generated automatically each time?

executor/workflow_manager_sfn.go
rgarcia: the use of stderr is born out of similar behavior in `gearcmd`. The theory behind that behavior is that failures are more often than not unexpected, in which case having easy access to stderr is useful for quickly debugging

mohit: so the idea is that in case of expected errors the `error.code` and `message` would be on `stdout` and the `stderr` would be the error lines? I'm fine with where this is right now, but having the full stderr for workers seems a little excessive.

executor/workflow_manager_sfn_test.go
```diff
@@ -44,3 +65,365 @@ func TestStateMachineName(t *testing.T) {
...
+			ExecutionArn: aws.String(sfnExecutionARN),
+		}).
+		Return(&sfn.DescribeExecutionOutput{
+			Status: aws.String(sfn.ExecutionStatusAborted),
+		}, nil)
+	require.NoError(t, c.manager.CancelWorkflow(workflow, reason))
+	assert.Equal(t, initialStatus, workflow.Status)
+	assert.Equal(t, reason, workflow.StatusReason)
+
+	t.Log("Verify execution is stopped and status reason is updated.")
```
mohit: `Verify failed execution status reason is updated`?

executor/workflow_manager_sfn_test.go
```diff
@@ -44,3 +65,365 @@ func TestStateMachineName(t *testing.T) {
...
+				jobFailedEvent,
+			}}, true)
+		})
+
+	require.NoError(t, c.manager.UpdateWorkflowStatus(workflow))
+	assert.Equal(t, models.WorkflowStatusFailed, workflow.Status)
+	require.Len(t, workflow.Jobs, 1)
+	assertBasicJobData(t, workflow.Jobs[0])
+	assert.Equal(t, models.JobStatusFailed, workflow.Jobs[0].Status)
+	assert.Equal(t, "line4\nline5\nline6", workflow.Jobs[0].StatusReason)
```
mohit: 💯

executor/workflow_manager_sfn.go
kofi-clever: yeah, i like the idea of having a snapshot of the stderr logs available for quick reference in the job (which also won't fall victim to log queue delays). we can always limit this further, but at least having the last line is helpful.

executor/workflow_manager_sfn.go
kofi-clever: if `JobIsDone() == true`, then the job is left alone (the workflow status already gets updated earlier).
wanted to make sure i don't corrupt job status if the cancel somehow happens between one task succeeding and the task exiting or the next task entering. in that case, don't want to override the successful job's status.

but looking at the code again, currentState would be nil anyway if we process a TaskExited event, so maybe this is a non-issue except for the case where an ActivitySucceeded event is followed by an ExecutionAborted, but don't know how likely that is. can leave this out for now until we notice any issues crop up

executor/workflow_manager_sfn_test.go
kofi-clever: oh. thanks. blind copy-paste

mocks/mock_sfniface/mock_sfnapi.go
kofi-clever: so far, we've been checking in the generated mocks, but we can probably eventually just include the mock gen as part of `make test` to avoid having to check them in.

executor/workflow_manager_sfn.go
mohit: hmm I see. Let's keep an eye out for this during the load testing over the next week.

PR #67, opened 2017-10-10
use ddb projection expression for summaryOnly=true

mohit:
This reduces our dynamodb resource usage when pulling large number of workflows.

- [N/A] Run "make generate"


store/dynamodb/workflow.go
```diff
@@ -11,6 +12,32 @@ import (
...
+	"Workflow.namespace",
+	"Workflow.retries",
+	"Workflow.retryFor",
+	"Workflow.#S", // status
+	"Workflow.statusReason",
+	"Workflow.tags",
+
+	"Workflow.workflowDefinition.#N",
+	"Workflow.workflowDefinition.version",
+}
```
kofi-clever: glad we're doing this!

is there a good place to add a note so people know to keep this in sync with any new fields added to the WAG model?
or maybe to the PR template or something?

store/dynamodb/dynamodb_store.go
```diff
@@ -556,7 +556,7 @@ func (d DynamoDB) GetWorkflows(query *store.WorkflowQuery) ([]models.Workflow, s
 	} else {
 		dbQuery, err = ddbWorkflowSecondaryKeyWorkflowDefinitionCreatedAt{
 			WorkflowDefinitionName: query.DefinitionName,
-		}.ConstructQuery()
+		}.ConstructQuery(query.SummaryOnly)
```
rgarcia: nit but it seems like WorkflowQuery should be defined in the swagger.yml, since all other inputs to the `Store` interface are encoded there

It could also let us encode constraints on the query, e.g. default list of summary fields

mohit: Thanks I'll make this change as part of this PR. I'm not sure how to encode the default list of summary fields in the `swagger.yml` - would love some guidance.

rgarcia: You could create a swagger type `WorkflowSummary` containing the subset of the `Workflow` object you want. Then you could add/test a method to the `Store` interface specifically for returning workflow summaries

mohit:
@rgarcia moved workflowQuery to be a swagger generated model and also added `WorkflowSummary` as an object that is a nested type in `Workflow`. I've not added a new `store` interface method that returns `WorkflowSummary` since that seemed to be more adding much more confusion than I think is worthwhile right now.

swagger.yml
```diff
@@ -387,10 +399,6 @@ definitions:
         $ref: '#/definitions/WorkflowStatus'
       statusReason:
```
kofi-clever: what do you think about making `statusReason` a detail field and dropping from the summary? realised we haven't needed it so far except in the detailed workflow view and given that it often contains bulky log lines, worth moving over to avoid loading excess data in the client (and to cut down on data loaded from dynamo).

swagger.yml
```diff
@@ -371,6 +373,16 @@ definitions:
       - "step-functions"
 
   Workflow:
+    allOf:
```
kofi-clever: 🎉

store/tests/tests.go
```diff
@@ -283,25 +288,34 @@ func GetWorkflowsSummaryOnly(s store.Store, t *testing.T) func(t *testing.T) {
...
 			Version: definition.Version,
 		}
 		require.Equal(t, definitionSummary, workflows[0].WorkflowDefinition)
 
+		// All fields of WorkflowSummary should be present
+		wsType := reflect.TypeOf(models.WorkflowSummary{})
+		workflowVal := reflect.ValueOf(workflows[0])
+		for i := 0; i < wsType.NumField(); i++ {
+			name := wsType.Field(i).Name
+			fmt.Println(i, workflowVal.FieldByName(name).String())
```
rgarcia: fmt.Println

store/tests/tests.go
```diff
@@ -283,25 +288,34 @@ func GetWorkflowsSummaryOnly(s store.Store, t *testing.T) func(t *testing.T) {
...
+			assert.NotNil(t, workflowVal.FieldByName(name).String())
```
rgarcia: 👏

swagger.yml
rgarcia: 🤘

store/tests/tests.go
mohit: 🤦‍♂️

swagger.yml
mohit: https://github.com/Clever/workflow-manager/pull/67/commits/0cf8a717b6a78054bffd1fc5ff63fa345df95f45

store/tests/tests.go
mohit: needed a `NotEmpty` also: https://github.com/Clever/workflow-manager/pull/67/commits/0cf8a717b6a78054bffd1fc5ff63fa345df95f45

mohit:
cool think this is ready for a 👍, let me know if there is anything else I should check.

PR #70, opened 2017-10-12
cache DescribeStateMachine calls

rgarcia:
https://clever.atlassian.net/browse/INFRA-2580

AWS rate limits DescribeStateMachine with a token bucket that refills at a rate of 1 per second. We will hit this pretty quickly under normal use.

This PR caches calls to that API method using an LRU cache w/ space for 1000 state machines. State machines are immutable and not that big, so caching this method doesn't present a huge risk.

### DIFF
--- a/executor/sfncache/sfn.go
+++ b/executor/sfncache/sfn.go
@@ -1,0 +1,39 @@
+package sfncache
+
+import (
+	"github.com/aws/aws-sdk-go/service/sfn"
+	"github.com/aws/aws-sdk-go/service/sfn/sfniface"
+	lru "github.com/hashicorp/golang-lru"
+)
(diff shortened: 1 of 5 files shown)


executor/sfncache/sfn.go
```diff
@@ -0,0 +1,39 @@
+package sfncache
+
+import (
+	"github.com/aws/aws-sdk-go/service/sfn"
+	"github.com/aws/aws-sdk-go/service/sfn/sfniface"
+	lru "github.com/hashicorp/golang-lru"
+)
+
+type sfnCache struct {
```
mohit: 💯 TIL overriding methods of an interface.

PR #71, opened 2017-10-12
generate VERSION for github-release on make generate

mohit:
Everytime `make generate` is run the `VERSION` file is updated with the swagger client version which can then be used by the `github-release` script to automatically tag the release. 

- [NA] Update swagger.yml version
- [ * ] Run "make generate"

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,3 +1,2 @@
 - [ ] Update swagger.yml version
 - [ ] Run "make generate"
-- [ ] After merging, add a github tag to your merge commit
--- a/circle.yml
+++ b/circle.yml
@@ -30,5 +30,8 @@ deployment:
       - $HOME/ci-scripts/circleci/dapple-deploy $DAPPLE_URL $DAPPLE_USER $DAPPLE_PASS $APP_NAME dev-infra confirm-then-deploy
       - $HOME/ci-scripts/circleci/dapple-deploy $DAPPLE_URL $DAPPLE_USER $DAPPLE_PASS $APP_NAME production confirm-then-deploy
       - $HOME/ci-scripts/circleci/npm-publish $NPM_TOKEN gen-js/
+      # create a VERSION file to push a github release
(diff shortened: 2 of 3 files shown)


Makefile
```diff
@@ -32,9 +32,13 @@ run-docker:
...
 	--env-file=<(echo -e $(_ARKLOC_ENV_FILE)) clever/workflow-manager:569f2dc
 
-generate: wag-generate-deps
+generate: wag-generate-deps VERSION
 	$(call wag-generate,./swagger.yml,$(PKG))
 
+VERSION: ./swagger.yml
+	@cat ./swagger.yml | grep "^  version:" | cut -d":" -f2 | tr -d " " > ./VERSION
+
+
```
rgarcia: extra newline

VERSION
```diff
@@ -0,0 +1 @@
+0.6.3
```
rgarcia: could this be gitignored?

Makefile
```diff
@@ -32,9 +32,13 @@ run-docker:
 	@docker run \
 	--env-file=<(echo -e $(_ARKLOC_ENV_FILE)) clever/workflow-manager:569f2dc
 
-generate: wag-generate-deps
+generate: wag-generate-deps VERSION
 	$(call wag-generate,./swagger.yml,$(PKG))
 
+VERSION: ./swagger.yml
+	@cat ./swagger.yml | grep "^  version:" | cut -d":" -f2 | tr -d " " > ./VERSION
```
rgarcia: since we're doing this to satisfy the github-release ci script, it's worth considering just running this command in CI prior to invoking the github release script

mohit: hmm, I added it this way so that we update the `VERSION` file only if you run `make generate`. My thinking was that this would mean that if you forget to `make generate` after updating the version in the `swagger.yml` file we would not generate a new VERSION file. 

Happy to change this to be a CI only thing though

mohit:
@rgarcia only creates the VERSION file in ci.

rgarcia:
🎉  if this works well we should add it to the wag template

mohit:
it works. Will follow up with a ticket for wag.

PR #73, opened 2017-10-13
POC an attempt at event processing tests

rgarcia:
- [ ] Run "make generate"


mohit:
@rgarcia for now it seems that we're using gomock to test similar flows. We should close out this PR or decide if we want to merge it.

PR #74, opened 2017-10-13
Fix job update logic for execution-level events

kofi-clever:
Make sure we update the last-seen job when we encounter an execution event (which wouldn't have a `PreviousEventId` value).

**Testing:**
- Confirmed that this properly sets the status and status reason on the last job if the execution fails. e.g. [hubble/87dc43fd-fc1b-4ccb-bdf5-acb8cc4b106c](https://clever-dev--hubble.int.clever.com/multiverse:master/workflows/w/87dc43fd-fc1b-4ccb-bdf5-acb8cc4b106c/jobs)
- (Compare with [hubble/28c60677-3da2-4199-bb6a-85e198c67ab4](https://clever-dev--hubble.int.clever.com/multiverse:master/workflows/apps/5674595f800a410001000004/w/28c60677-3da2-4199-bb6a-85e198c67ab4/jobs))

**Rollout:**
- [N/A] Run "make generate"

### DIFF
--- a/executor/workflow_manager_sfn.go
+++ b/executor/workflow_manager_sfn.go
@@ -390,11 +390,12 @@ func (wm *SFNWorkflowManager) UpdateWorkflowStatus(workflow *models.Workflow) er
 		// 3) stop paging once we get to to the smallest job ID (aka event ID) that is still pending
 		for _, evt := range historyOutput.Events {
 			job := eventToJob(evt)
-			if job == nil {
-				continue
-			}
 			switch aws.StringValue(evt.Type) {
 			case sfn.HistoryEventTypeTaskStateEntered:
+				if job == nil {
+					continue
+				}
+
 				// event IDs start at 1 and are only unique to the execution, so this might not be ideal
 				job.ID = fmt.Sprintf("%d", aws.Int64Value(evt.Id))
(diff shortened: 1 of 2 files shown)


executor/workflow_manager_sfn.go
```diff
@@ -458,14 +471,26 @@ func (wm *SFNWorkflowManager) UpdateWorkflowStatus(workflow *models.Workflow) er
...
+
 				job.Status = models.JobStatusAbortedByUser
 				job.StoppedAt = strfmt.DateTime(aws.TimeValue(evt.Timestamp))
 				if details := evt.ExecutionAbortedEventDetails; details != nil {
 					job.StatusReason = aws.StringValue(details.Cause)
 				}
 			case sfn.HistoryEventTypeExecutionFailed:
+				if job == nil {
+					// Execution-level event - update last seen job.
+					job = jobs[len(jobs)-1]
```
rgarcia: can this go in `eventToJob` instead? That's where I'd like to encode all of the logic around mapping events to jobs

kofi-clever: oh that makes sense - https://github.com/Clever/workflow-manager/pull/74/commits/3b13cb55b98e28464cfcbe5ff5eec8b63bdaaa89

PR #76, opened 2017-10-18
cancellation of failed workflows

mohit:
We allow Cancellation of Failed Workflows. However in the current implementation this only worked while SFN kept the execution ID in history.

If a Workflow is complete, we should allow marking a Workflow as Cancelled.

The logic here is a bit confusing but simple, any ideas on how to write it better welcome. 

- [ NA ] Update swagger.yml version
- [ NA ] Run "make generate"

### DIFF
--- a/executor/workflow_manager_sfn.go
+++ b/executor/workflow_manager_sfn.go
@@ -247,26 +247,34 @@ func (wm *SFNWorkflowManager) RetryWorkflow(ogWorkflow models.Workflow, startAt,
 }
 
 func (wm *SFNWorkflowManager) CancelWorkflow(workflow *models.Workflow, reason string) error {
-	// cancel execution
-	wd := workflow.WorkflowDefinition
-	execARN := wm.executionARN(workflow, wd)
-	_, err := wm.sfnapi.StopExecution(&sfn.StopExecutionInput{
-		ExecutionArn: aws.String(execARN),
-		Cause:        aws.String(reason),
-		// Error: aws.String(""), // TODO: Can we use this? "An arbitrary error code that identifies the cause of the termination."
-	})
-	if err != nil {
-		return err
+	if workflow.Status == models.WorkflowStatusSucceeded {
+		return fmt.Errorf("Cancellation not allowed. Workflow %s is %s", workflow.ID, workflow.Status)
 	}
 
+	// attempt to describe execution
+	wd := workflow.WorkflowDefinition
+	execARN := wm.executionARN(workflow, wd)
 	describeOutput, err := wm.sfnapi.DescribeExecution(&sfn.DescribeExecutionInput{
 		ExecutionArn: aws.String(execARN),
 	})
 	if err != nil {
 		return err
 	}
-
 	status := sfnStatusToWorkflowStatus(*describeOutput.Status)
+
+	if _, err = wm.sfnapi.StopExecution(&sfn.StopExecutionInput{
+		ExecutionArn: aws.String(execARN),
+		Cause:        aws.String(reason),
+		// Error: aws.String(""), // TODO: Can we use this? "An arbitrary error code that identifies the cause of the termination."
+	}); err != nil {
+		aerr, ok := err.(awserr.Error)
+		// AWS returns a DoesNotExist error when cancelling a failed workflow
+		executionFailedErr := (ok && aerr.Code() == sfn.ErrCodeExecutionDoesNotExist && status == models.WorkflowStatusFailed)
+		if !executionFailedErr {
+			return err
+		}
+	}
+
 	if status == models.WorkflowStatusFailed {
 		// SFN executions can't be moved from the failed state to the aborted state, so if the user
 		// cancels a failed workflow, we update the status in workflow-manager directly to preserve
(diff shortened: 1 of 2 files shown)


executor/workflow_manager_sfn.go
```diff
@@ -247,26 +247,38 @@ func (wm *SFNWorkflowManager) RetryWorkflow(ogWorkflow models.Workflow, startAt,
...
+		describeOutput, err := wm.sfnapi.DescribeExecution(&sfn.DescribeExecutionInput{
+			ExecutionArn: aws.String(execARN),
+		})
+		if err != nil {
+			return err
+		}
+		status := sfnStatusToWorkflowStatus(*describeOutput.Status)
 	}
 
-	status := sfnStatusToWorkflowStatus(*describeOutput.Status)
```
rgarcia: potentially cleaner way to do this (that also checks the cast to `awserr.Error`):
```go
	if err != nil {
		aerr, ok := err.(awserr.Error)
		// ignore cases where Execution is not found after Workflow execution has completed
		// this happens if we try to cancel a failed workflow after SFN's TTL for executions (
		executionTTLd := (ok && aerr.Code() == sfn.ErrCodeExecutionDoesNotExist && resources.WorkflowIsDone(workflow))
		if !executionTTLd {
			return err
		}
	} else {
		// if cancellation worked check current status of Workflow
		describeOutput, err := wm.sfnapi.DescribeExecution(&sfn.DescribeExecutionInput{
			ExecutionArn: aws.String(execARN),
		})
		if err != nil {
			return err
		}
		if sfnStatusToWorkflowStatus(*describeOutput.Status) == models.WorkflowStatusFailed {
			// SFN executions can't be moved from the failed state to the aborted state, so if the user
			// cancels a failed workflow, we update the status in workflow-manager directly to preserve
			// user intent.
			workflow.Status = models.WorkflowStatusCancelled
		}
	}
```

rgarcia: also do we know what the TTL is for executions?

mohit: The TTL is supposed to be 90 days and there is something else going on here. One of them might have been manually deleted or just failed to start. A few others I can see in the UI but am not able to access them using the CLI. 

https://github.com/Clever/workflow-manager/pull/76/commits/b1f3be98de38c341cb874c5a64d0c8d6f62776b3

mohit:
Leaving out the TTL'ed issue from here and going to make sure we actually have a case to cancel things after 90 days. For now this should help us cancel workflows that exist but are failed.

executor/workflow_manager_sfn.go
```diff
@@ -247,26 +247,35 @@ func (wm *SFNWorkflowManager) RetryWorkflow(ogWorkflow models.Workflow, startAt,
 }
 
 func (wm *SFNWorkflowManager) CancelWorkflow(workflow *models.Workflow, reason string) error {
-	// cancel execution
+	if workflow.Status == models.WorkflowStatusSucceeded {
+		return fmt.Errorf("Cancellation not allowed. Workflow %s is %s", workflow.ID, workflow.Status)
+	}
+
+	status := workflow.Status
```
rgarcia: I'm pretty sure this declaration is unnecessary and can be moved down to the line

```
status := sfnStatusToWorkflowStatus(*describeOutput.Status)
```

executor/workflow_manager_sfn.go
```diff
@@ -247,26 +247,35 @@ func (wm *SFNWorkflowManager) RetryWorkflow(ogWorkflow models.Workflow, startAt,
...
+		executionFailedErr := (ok && aerr.Code() == sfn.ErrCodeExecutionDoesNotExist && status == models.WorkflowStatusFailed)
+		if !executionFailedErr {
+			return err
+		}
 	}
 
-	status := sfnStatusToWorkflowStatus(*describeOutput.Status)
 	if status == models.WorkflowStatusFailed {
 		// SFN executions can't be moved from the failed state to the aborted state, so if the user
 		// cancels a failed workflow, we update the status in workflow-manager directly to preserve
```
rgarcia: sidenote, but I'm not sure I agree with this setting of the workflow status...it seems to hide the fact that the worfklow failed, which is more important than the fact that someone tried to cancel it after it failed.

What do you think of cancellation failing unless the execution is in progress? If the execution finished (either succeeded or failed), I don't see the point in processing a cancellation.

mohit: This is currently required for the way we are managing success/failure in Hubble and for multiverse. I'll take up exploring other options in this in the next couple of weeks and at that point we can reconsider this behavior. 

Summary of use case where we `cancel` a failed workflow. It allows us to mark that we:

* don't care about the failure anymore
* we're retried it and therefore ^

This is used to differentiate "failed" therefore take action from "cancelled" ignore. I don't think it's a good model, but don't want to break the hubble UI for this right now.

executor/workflow_manager_sfn.go
mohit: 🙇 https://github.com/Clever/workflow-manager/pull/76/commits/ecc0fb7d2b2eb1a5868aed03502278f0d22bacff

executor/workflow_manager_sfn.go
rgarcia: I agree it's not a great model.. seems like a workaround for the lack of a real mechanism to mark / ignore a failure. Seems worth some thinking about but not urgent!

PR #78, opened 2017-10-18
Add support for Choice, Fail states

rgarcia:
https://clever.atlassian.net/browse/INFRA-2589

Adds fields required to support these state types. This brings our support of states types to

- [x] Pass 
- [x] Task
- [x] Choice
- [ ] Wait: would require adding ”Seconds”, “SecondsPath”, “Timestamp”, or “TimestampPath” fields to the SLState type (easy)
- [x] Succeed
- [x] Fail
- [ ] Parallel: would require adding "Branches" array to the SLState type, which is an array of state-machine like objects


rgarcia:
cc @mohit: I removed calls to inactivestates in 585fa08, but [I am still not sure what the reasoning was for having those calls in the first place](https://en.wikipedia.org/wiki/Wikipedia:Chesterton%27s_fence), so maybe you should review that

alsmola:
Not 100% sure what I'm looking at, but don't see anything obviously wrong so :+1:

executor/workflow_manager_sfn.go
```diff
@@ -214,10 +214,6 @@ func (wm *SFNWorkflowManager) RetryWorkflow(ogWorkflow models.Workflow, startAt,
 	// modify the StateMachine with the custom StartState by making a new WorkflowDefinition (no pointer copy)
 	newDef := resources.CopyWorkflowDefinition(*ogWorkflow.WorkflowDefinition)
 	newDef.StateMachine.StartAt = startAt
-	if err := resources.RemoveInactiveStates(newDef.StateMachine); err != nil {
```
mohit: SFN will fail  to create a state machine that has states that have no active path to. All states must have a possible path. RemoveInactiveStates ensures that we will submit valid StateMachines.

executor/workflow_manager_sfn.go
```diff
@@ -362,6 +358,9 @@ func (wm *SFNWorkflowManager) UpdateWorkflowStatus(workflow *models.Workflow) er
 		case sfn.HistoryEventTypeExecutionStarted:
 			// very first event for an execution, so there are no jobs yet
 			return nil
+		case sfn.HistoryEventTypeChoiceStateEntered, sfn.HistoryEventTypeChoiceStateExited, sfn.HistoryEventTypeFailStateEntered, sfn.HistoryEventTypeSucceedStateEntered, sfn.HistoryEventTypeSucceedStateExited, sfn.HistoryEventTypePassStateEntered, sfn.HistoryEventTypePassStateExited, sfn.HistoryEventTypeParallelStateEntered, sfn.HistoryEventTypeParallelStateExited, sfn.HistoryEventTypeWaitStateEntered, sfn.HistoryEventTypeWaitStateExited:
```
mohit: nit: break into multiple lines?

handler.go
```diff
@@ -274,29 +274,11 @@ func (h Handler) ResumeWorkflowByID(ctx context.Context, input *models.ResumeWor
...
 	}
 
-	for _, s := range req.StateMachine.States {
-		if s.Type != "" && s.Type != models.SLStateTypeTask {
-			return nil, fmt.Errorf("Only States of `type=Task` are supported")
-		}
-	}
-
-	// ensure all states are defined and have a transition path
-	numStates := len(req.StateMachine.States)
```
mohit: This validates a StateMachine

executor/workflow_manager_sfn.go
rgarcia: what if instead we validate state machines in `ark-config`, perhaps via the tool they provide? I feel like we're taking on a pretty big responsibility (and will have a hard time) nailing validation of state machines here.

The benefit of validating here also seems pretty small... the user will already get feedback that the state machine is invalid when they go to submit the first execution

executor/workflow_manager_sfn.go
rgarcia: 8244261

executor/workflow_manager_sfn.go
mohit: We can not support retry or replay without validation of states and removing the ones that are not in the retry. Summary: all state machines where startAt is not set at the beginning of the graph are considered invalid. The removeInactiveStates is a poly fill for that

rgarcia: 😞  filed a feature request w/ AWS to support executing a state machine from a different StartAt state

rgarcia:
@mohit can you review 42e762aa5ae6de231d54d892667d7ee75c0d659b

I added back and sort of re-wrote the validation logic

PR #81, opened 2017-10-25
remove batch code

rgarcia:
https://clever.atlassian.net/browse/INFRA-2576


rgarcia:
only references to batch are now in kvconfig... @mohit not familiar with these, can you advise?

handler.go
```diff
@@ -121,8 +126,12 @@ func (h Handler) PutStateResource(ctx context.Context, i *models.PutStateResourc
 		}
 	}
 
-	stateResource := resources.NewBatchResource(
-		i.NewStateResource.Name, i.NewStateResource.Namespace, i.NewStateResource.URI)
+	stateResource := &models.StateResource{
```
mohit: lastUpdated?

handler.go
```diff
@@ -101,7 +101,12 @@ func (h Handler) GetWorkflowDefinitionByNameAndVersion(ctx context.Context, inpu
 
 // PostStateResource creates a new state resource
 func (h Handler) PostStateResource(ctx context.Context, i *models.NewStateResource) (*models.StateResource, error) {
-	stateResource := resources.NewBatchResource(i.Name, i.Namespace, i.URI)
+	stateResource := &models.StateResource{
+		Name:      i.Name,
+		Namespace: i.Namespace,
+		URI:       i.URI,
+		Type:      models.StateResourceTypeActivityARN,
+	}
```
mohit: lastUpdated

mohit:
We should keep the `log_helpers` files but can remove the `batch only` log routes. I marked them as such yesterday expecting to delete or move them over to step functions. Removing them is good for now, we can re-add them only if we need additional logs.

kvconfig.yml
```diff
@@ -2,28 +2,6 @@
 routes:
 
   # matchers for tracking Jobs/Workflows; prefix with `workflows.`
```
mohit: this line is no longer useful.

handler.go
rgarcia: a8ae337

handler.go
rgarcia: a8ae337

PR #82, opened 2017-10-25
[Proposal] Start parsing details for Choice and Succeed tasks for client visibility

kofi-clever:
Would like to include details about which branch of the workflow was taken, so that it's clearer to users (and the client doesn't have to extrapolate).

**Basically go from this (no job data past last Task):**
https://clever-dev--hubble.int.clever.com/multiverse:master/workflows/w/77e8b1f0-a2e7-4b28-8794-438ce8a0cb23/jobs
Screenshot from Gyazo

**to this (Choice and Succeed step details included):**
https://clever-dev--hubble.int.clever.com/multiverse:master/workflows/w/1ec091c2-3e88-4087-8ffa-e4f9c47c6851/jobs
Screenshot from Gyazo

- [N/A] Run "make generate"

### DIFF
--- a/executor/workflow_manager_sfn.go
+++ b/executor/workflow_manager_sfn.go
@@ -369,15 +369,13 @@ func (wm *SFNWorkflowManager) UpdateWorkflowStatus(workflow *models.Workflow) er
 		case sfn.HistoryEventTypeExecutionStarted:
 			// very first event for an execution, so there are no jobs yet
 			return nil
-		case sfn.HistoryEventTypeChoiceStateEntered, sfn.HistoryEventTypeChoiceStateExited,
-			sfn.HistoryEventTypeSucceedStateEntered, sfn.HistoryEventTypeSucceedStateExited,
-			sfn.HistoryEventTypePassStateEntered, sfn.HistoryEventTypePassStateExited,
+		case sfn.HistoryEventTypePassStateEntered, sfn.HistoryEventTypePassStateExited,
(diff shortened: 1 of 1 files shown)


kofi-clever:
@rgarcia + @mohit - thoughts on parsing some of the non-task states for clarity?

rgarcia:
Nice! I fully support mapping more of the execution history into the `Jobs` array. Nit, but we should remove language / comments like `a job is created when a task state is entered` and `only Task states have jobs`.

Long term I think we will need to rethink the `Jobs` array, since I don't think it is a great data model for representing the execution of a state machine, which is more tree-like than array-like. Food for thought: how would the jobs array / Hubble present a state machine with parallel execution paths?

mohit:
+1 on shipping this after changing some of the comments terminology. I'm adding a spec task for expected user interaction for jobs (e.g. can be a tree, attempts, state information...). Once we have some requirements, we can look into a better data model for Jobs.

mohit:
@rgarcia does https://github.com/Clever/workflow-manager/pull/82/commits/90d0069989cc13d56dc50be73b168265feb46944 seem good?

@kofi-clever is that okay if I go ahead and merge this? Any further changes/work required to support this on the hubble end?

rgarcia:
Looks good to me

kofi-clever:
@mohit - sorry for the delay! back in the office today. this is good to go if you want to merge. hubble changes are already in place to support this

PR #84, opened 2017-10-26
log / route count of calls to GetExecutionHistory

rgarcia:
Adds an implementation of sfniface.SFNAPI that intercepts calls to `GetExecutionHistoryPagesWithContext` and increments a counter. Logs the value of this counter and routes it to signalfx.

This should let us monitor our total request rate to this method, and also give us a good idea of how effective our strategies are around controlling our request rate.

It might also serve as template for doing this elsewhere in our code where we hit AWS API limits (I'm looking at you, ecs-metrics 😜 )

### DIFF
--- a/executor/sfncounter/cumulativebucket.go
+++ b/executor/sfncounter/cumulativebucket.go
@@ -1,0 +1,97 @@
+// this file borrows from https://github.com/signalfx/golib
+package sfncounter
+
+import (
+	"math"
+	"sync/atomic"
+	"time"
+
+	"github.com/signalfx/golib/datapoint"
+)
+
+type atomicFloat struct {
+	bits uint64
+}
+
+func (a *atomicFloat) Get() float64 {
+	return math.Float64frombits(atomic.LoadUint64(&a.bits))
+}
+
+func (a *atomicFloat) Add(f float64) {
+	for {
+		oldValue := atomic.LoadUint64(&a.bits)
+		newValue := math.Float64bits(math.Float64frombits(oldValue) + f)
+		if atomic.CompareAndSwapUint64(&a.bits, oldValue, newValue) {
+			break
+		}
+	}
+}
+
+// Result is a cumulated result of items that can be added to a CumulativeBucket at once
+type Result struct {
+	Count        int64
+	Sum          int64
+	SumOfSquares float64
+}
+
+// Add a single number to the bucket.  This does not use atomic operations and is not thread safe,
+// but adding a finished Result into a CumulativeBucket is thread safe.
+func (r *Result) Add(val int64) {
+	r.Count++
+	r.Sum += val
+	// Sum of squares tends to roll over pretty easily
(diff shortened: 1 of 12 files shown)


executor/sfncounter/cumulativebucket.go
```diff
@@ -0,0 +1,97 @@
...
+		return []*datapoint.Datapoint{}
+	}
+	return []*datapoint.Datapoint{
+		CumulativeP(b.MetricName+".count", b.Dimensions, &b.count),
+		CumulativeP(b.MetricName+".sum", b.Dimensions, &b.sum),
+		CumulativeF(b.MetricName+".sumsquare", b.Dimensions, b.sumOfSquares.Get()),
+	}
+}
+
+// CumulativeP creates a SignalFx cumulative counter for integer values from a pointer that is loaded atomically.
```
mohit: We can remove the Signalfx mentions from these comments?

kvconfig.yml
```diff
@@ -30,3 +30,12 @@ routes:
...
       dimensions: []
+
+  get-execution-history-count:
+    matchers:
+      title: ["get-execution-history-count"]
+    output:
+      type: "alerts"
+      series: "workflow-manager.get-execution-history-count"
+      stat_type: "counter"
+      dimensions: []
```
mohit: Would it be useful/simple to add a dimension for execution-id or state-machine-name?

executor/sfncounter/cumulativebucket.go
rgarcia: this file is a literal copy-paste of the `cumulativebucket.go` file. We could finesse the content of it, but then the line between copy-paste and "fork with substantive modifications" starts to blur.

mohit: Hmm, I was thinking of this more as a inspired by, since it seems to me we should not really depend on the signalfx library here at all. I don't think this should be a blocker for now, but would like to know if you think we should keep this interaction with signalfx library long term.

kvconfig.yml
rgarcia: the space of execution ID would be pretty huge, but state machine could be useful...

The complexity increases once we need to count by state machine, though, so I'd rather get a global count for now and if we need to start digging in by state machine add that down the road

executor/sfncounter/cumulativebucket.go
rgarcia: yeah I agree we shouldn't pull in the signalfx client, so that's why I copy-pasted this. The only dependency I added is on the `datapoint` sub package, which seems reasonable given that we're tracking a counter

mohit: Thanks for the explanation. I agree this the `datapoint` dependency is fine iuf it's isolated to this package.

I was also getting confused between `sfncounter` and `sfx` for signalfx 🤦‍♂️ (my motivation for arguing to remove references to signalfx). But now I get it :)

PR #88, opened 2017-10-31
Prepend defaultRetry to conform to spec

mohit:
From https://states-language.net/spec.html#error-names

> The reserved name “States.ALL” in a Retrier’s “ErrorEquals” field is a
wild-card and matches any Error Name. *Such a value MUST appear alone in
the “ErrorEquals” array and MUST appear in the last Retrier in the
“Retry” array.*

- [ na ] Update swagger.yml version
- [ na ] Run "make generate"


mohit:
cc: @bgveenstra @rgarcia

rgarcia:
nice catch!

PR #89, opened 2017-11-01
tweak count metrics

rgarcia:
make the routing here consistent with https://github.com/Clever/ecs-metrics/pull/43


rgarcia:
@mohit counters now feeding a single metric series `aws-sdk-go.counter` with dimensions for `aws-service` and `aws-operation`

this should let us create a chart in sfx that shows a global view, but also let us drill down by dimension

PR #90, opened 2017-11-02
Remove workflow sfn create fail

mohit:
Progress on INFRA-2607: Job history occasionally missing from workflows

I was able to reproduce a case where we failed to startExecution due to rate limiting and then status updates kept failing since we could not find the execution. This change attempts to make `CreateWorkflow` to be more transactional. 

TODO: add tests for this condition. ✅ 

- [ na ] Update swagger.yml version
- [ na ] Run "make generate"

### DIFF
--- a/executor/workflow_manager_sfn.go
+++ b/executor/workflow_manager_sfn.go
@@ -211,10 +211,10 @@ func (wm *SFNWorkflowManager) CreateWorkflow(wd models.WorkflowDefinition,
 		// since we failed to start execution, remove Workflow from store
 		if delErr := wm.store.DeleteWorkflowByID(workflow.ID); delErr != nil {
 			log.ErrorD("create-workflow", logger.M{
-				"id":            workflow.ID,
-				"workflow-name": workflow.WorkflowDefinition.Name,
-				"message":       "failed to delete stray workflow",
-				"error":         fmt.Sprintf("SFNError: %s;StoreError: %s", err, delErr),
+				"id": workflow.ID,
+				"workflow-definition-name": workflow.WorkflowDefinition.Name,
+				"message":                  "failed to delete stray workflow",
+				"error":                    fmt.Sprintf("SFNError: %s;StoreError: %s", err, delErr),
 			})
 		}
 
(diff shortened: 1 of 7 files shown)


mohit:
cc: @kofi-clever

executor/workflow_manager_sfn.go
```diff
@@ -203,6 +208,16 @@ func (wm *SFNWorkflowManager) CreateWorkflow(wd models.WorkflowDefinition, input
 	// submit an execution using input, set execution name == our workflow GUID
 	err = wm.startExecution(describeOutput.StateMachineArn, workflow.ID, input)
 	if err != nil {
+		// since we failed to start execution, remove Workflow from store
+		if delErr := wm.store.DeleteWorkflowByID(workflow.ID); delErr != nil {
+			log.ErrorD("create-workflow", logger.M{
+				"id":            workflow.ID,
+				"workflow-name": workflow.WorkflowDefinition.Name,
```
rgarcia: nit: `workflow-definition-name`?

executor/workflow_manager_sfn_test.go
```diff
@@ -127,6 +127,78 @@ func TestStateMachineWithDefaultRetriers(t *testing.T) {
...
+	savedWorkflow, err := c.store.GetWorkflowByID(workflow.ID)
+	assert.Nil(t, err)
+	assert.Equal(t, workflow.CreatedAt.String(), savedWorkflow.CreatedAt.String())
+	assert.Equal(t, workflow.ID, savedWorkflow.ID)
+	pendingIDs, err := c.store.GetPendingWorkflowIDs()
+	assert.Nil(t, err)
+	assert.Equal(t, 1, len(pendingIDs))
+	assert.Equal(t, workflow.ID, pendingIDs[0]) // current workflow is pending
+
+	t.Log("CreateWorkflow deletes workflow on StartExecution failure")
```
rgarcia: consider using `t.Run()` to logically group sub tests. I've found it hard to debug test functions that pack lots of logically different tests with their own mock expectations into the same `Test...` scope. E.g., defer'd mock assertions failing require looking at all of the `EXPECT`s. `t.Run()` lets you do independent setup/teardown of mocks and will isolate failures of `defer`d assertions to specific, named subtests.

mohit: https://github.com/Clever/workflow-manager/pull/90/commits/6e8f91232cacf29db43643c1c5d8ce73a4a9d4e4

executor/workflow_manager_sfn.go
mohit: 👍 https://github.com/Clever/workflow-manager/pull/90/commits/616905adc0ada5f094556552921193d6855174a1

PR #91, opened 2017-11-08
consistent read in queries

rgarcia:
make sure the update loop (and other queries) don't suffer from eventual consistency


rgarcia:
blugh nevermind

>When you put or delete items in a table, the global secondary indexes on that table are updated in an eventually consistent fashion. Changes to the table data are propagated to the global secondary indexes within a fraction of a second, under normal conditions. However, in some unlikely failure scenarios, longer propagation delays might occur. Because of this, your applications need to anticipate and handle situations where a query on a global secondary index returns results that are not up-to-date.

from http://docs.aws.amazon.com/amazondynamodb/latest/developerguide/GSI.html

PR #93, opened 2017-11-09
Record the `stoppedAt` time when marking a workflow as done.

kofi-clever:
Useful in the client for sorting by recently completed and for exposing workflow duration without having to fetch job details.

- [x] Run "make generate"


executor/workflow_manager_sfn.go
```diff
@@ -381,6 +381,9 @@ func (wm *SFNWorkflowManager) UpdateWorkflowSummary(workflow *models.Workflow) e
 
 	workflow.LastUpdated = strfmt.DateTime(time.Now())
 	workflow.Status = sfnStatusToWorkflowStatus(*describeOutput.Status)
+	if resources.WorkflowIsDone(workflow) {
```
mohit: I don't think we need to run this check here afaik. We can just check if `StopDate` is set?

kofi-clever:
oh, thanks! forgot i hadn't sent this out 😄

PR #94, opened 2017-11-09
[INFRA-2606] better handling for job and workflow timeouts

bgveenstra:
**Jira**: [INFRA-2606](https://clever.atlassian.net/projects/INFRA/issues/INFRA-2606)

**Overview**:
Sometimes, jobs or workflows time out. We currently treat timeouts as general failures, which has produced confusing or misleading hubble output. This PR lets workflow-manager keep job and workflow statuses updated and provide more informative status reasons when timeouts happen. Hubble can use these changes to make debugging easier!

**Job (activity) time out**:
Example: FAILED executions of our `activity-timed-out` state machine in the [AWS console](https://us-west-2.console.aws.amazon.com/states/home)

A job times out at <`ActivityStarted` time> + <`state.TimeoutSeconds`>.

When a job times out, it results in history events `ActivityTimedOut`.   We use this information to update the job's status reason in `UpdateWorkflowHistory`.

~A job timing out also results in the status of the execution being failed with output that includes an error `States.Timeout`.  We use this information to update the failed workflow's status reason in `UpdateWorkflowSummary`.~

execution history for an execution that failed when an activity timed out

**Workflow (execution) time out**:
Example: TIMED OUT executions of our `execution-timed-out` state machine in the [AWS console](https://us-west-2.console.aws.amazon.com/states/home)

When a workflow times out, it results the history event `ExecutionTimedOut`, and it changes the status of the execution to `Timed Out`.  We use this information to update the failed workflow's status reason in `UpdateWorkflowSummary`.

execution history for an execution that timed out

*To Do*:
- [x] Run "make generate"


resources/jobs.go
```diff
@@ -39,3 +39,11 @@ func JobStatusToInt(status models.JobStatus) int {
 		return 0
 	}
 }
+
+const (
```
mohit: These named as `StatusReasonJobTimedOut` and `StatusReasonWorkflowTimedOut` would be more aligned with the current naming scheme in the `resources` package.

executor/workflow_manager_sfn.go
```diff
@@ -381,7 +381,14 @@ func (wm *SFNWorkflowManager) UpdateWorkflowSummary(workflow *models.Workflow) e
 
 	workflow.LastUpdated = strfmt.DateTime(time.Now())
 	workflow.Status = sfnStatusToWorkflowStatus(*describeOutput.Status)
+	if *describeOutput.Status == sfn.ExecutionStatusTimedOut {
+		workflow.StatusReason = resources.AdditionalInfoExecutionTimedOut
+	}
 	workflow.Output = aws.StringValue(describeOutput.Output) // use for error or success
```
mohit: maybe move this to be after the StatusReason update blocks?

executor/workflow_manager_sfn.go
```diff
@@ -514,6 +524,20 @@ func (wm *SFNWorkflowManager) UpdateWorkflowHistory(workflow *models.Workflow) e
 						aws.StringValue(details.Error),
 					))
 				}
+			case sfn.HistoryEventTypeActivityTimedOut:
+				job.Status = models.JobStatusFailed
+				job.StoppedAt = strfmt.DateTime(aws.TimeValue(evt.Timestamp))
+				if details := evt.ActivityTimedOutEventDetails; details != nil {
+					// TODO: need more natural place to put error name...
```
mohit: This seems like a good place to put the error name. I'd suggest we always start with `resources.AdditionalInfoActivityTimeout:`. So it could be 

```golang
job.StatusReason = resources.AdditionalInfoActivityTimedOut

if details := evt.ActivityTimedOutEventDetails; details != nil {
    job.StatusReason = fmt.Sprintf("%s\n%s\n%s", job.StatusReason, aws.StringValue(details.Error), getLastFewLines(details.Cause))
}
```

executor/workflow_manager_sfn.go
```diff
@@ -528,6 +552,10 @@ func (wm *SFNWorkflowManager) UpdateWorkflowHistory(workflow *models.Workflow) e
 				if details := evt.ExecutionFailedEventDetails; details != nil {
 					if isActivityDoesntExistFailure(evt.ExecutionFailedEventDetails) {
 						job.StatusReason = "State resource does not exist"
+					} else if isActivityTimedOutFailureEvent(evt.ExecutionFailedEventDetails) {
+						// do not update job status reason -- it should already reflect the activity timeout
```
mohit: I think it's okay to just say here that we don't want to update job status reason since it would be already done by the ActivityTimedOut event.

executor/workflow_manager_sfn.go
mohit: hmm, not 💯 about this -> take it or leave it

executor/workflow_manager_sfn.go
```diff
@@ -569,6 +611,35 @@ func isActivityDoesntExistFailure(details *sfn.ExecutionFailedEventDetails) bool
 		strings.Contains(aws.StringValue(details.Cause), "Internal Error")
 }
 
+// isActivityTimedOutFailureEvent checks if an execution failed because an activity timed out,
+// based on the details of an 'execution failed' history event.
+func isActivityTimedOutFailureEvent(details *sfn.ExecutionFailedEventDetails) bool {
+	return aws.StringValue(details.Error) == "States.Timeout"
+}
+
+type TimedOutErrorOutput struct {
```
mohit: Isn't this already defined as `sfn.ActivityTimedOutEventDetails`: [aws-sdk-go](https://docs.aws.amazon.com/sdk-for-go/api/service/sfn/#ActivityTimedOutEventDetails)

mohit: We should be able to re-use the `isActivityTimedOutFailure` for checking in UpdateWorkflowSummary too. (I think)

bgveenstra: > Isn't this already defined as `sfn.ActivityTimedOutEventDetails`
I should fix the name.  This is actually here to help unmarshal the [output of a failed execution, which is just a string](https://docs.aws.amazon.com/sdk-for-go/api/service/sfn/#DescribeExecutionOutput).  
Unfortunately, this whole thing is moot.  [API documentation](http://docs.aws.amazon.com/step-functions/latest/apireference/API_DescribeExecution.html) says `output` is only sent if the execution succeeds, and we've confirmed by hitting the API. 

That means we have no way to figure out whether the execution failed due to an activity timeout unless we look at history events. (And we've so far avoided updating `workflow.Status` during `UpdateWorkflowSummary`)

Also the `omitempty` never made, my bad.

bgveenstra: > We should be able to re-use the `isActivityTimedOutFailure` for checking in `UpdateWorkflowSummary` too.

Initially, it was important to be able to get this information from the history events and from the execution output. Given the information above which shows that that's not possible, we can certainly remove the `isActivityTimedOutFailureStatus` function.

resources/jobs.go
bgveenstra: I avoided this because they become part of the `StatusReason`, not the entire reason.

executor/workflow_manager_sfn.go
bgveenstra: Made this change in 0c656df47ee0dc6884691eaefbd4dec8bff1c535

executor/workflow_manager_sfn.go
bgveenstra: sounds good, updated in f9f31eae2bac215cd76f70de2590fe59fdc4fecc

executor/workflow_manager_sfn.go
bgveenstra: Removed code that attempted to update workflow status reason based on non-existent output in d19d7dcc9a5a04541ecb883c84145c382960e7b5 and commented out related test in cd1e1e13e904a1686ed5c025b92a51b695fd0880

bgveenstra: Thanks for talking this over with me @mohit.  To document that decision - the benefit we gain from having the failed workflow status reason mention the activity time out is not worth the intertwining of code we would have to do to update based on history events. We're going to skip modifying the workflow status reason when an activity times out.

Final commit in removing that old code was 94f0738d389cde7d15e6c6b723e097a1c8d4a38b

executor/workflow_manager_sfn.go
```diff
@@ -381,7 +381,11 @@ func (wm *SFNWorkflowManager) UpdateWorkflowSummary(workflow *models.Workflow) e
 
 	workflow.LastUpdated = strfmt.DateTime(time.Now())
 	workflow.Status = sfnStatusToWorkflowStatus(*describeOutput.Status)
-	workflow.Output = aws.StringValue(describeOutput.Output) // use for error or success
+	if *describeOutput.Status == sfn.ExecutionStatusTimedOut {
+		workflow.StatusReason = resources.AdditionalInfoExecutionTimedOut
+	}
+	workflow.Output = aws.StringValue(describeOutput.Output) // use for error or success  (@todo actually this is only sent for success)
```
bgveenstra: I doubt we're relying on this having output for failure, but it doesn't. Maybe we should check.

resources/jobs.go
mohit: 👍 still using both Activity and Job wording. I think sticking with one would be better -> in this case since it's user facing I'm leaning towards Job and Workflows.

executor/workflow_manager_sfn_test.go
```diff
@@ -725,6 +725,148 @@ func TestUpdateWorkflowStatusExecutionNotFoundStopRetry(t *testing.T) {
...
+	ExecutionFailedEventDetails: &sfn.ExecutionFailedEventDetails{
+		Error: aws.String("States.Timeout"),
+	},
+}
+
+func assertTimedOutJobData(t *testing.T, job *models.Job) {
+	assert.Equal(t, models.JobStatusFailed, job.Status)
+	// not currently using the details in TestUpdateWorkflowStatusJobTimedOut test
+	// also this format does not match the output in workflow_manager_sfn
+	assert.Contains(t, job.StatusReason, "Job timed out")
```
mohit: use `resources.StatusReasonJobTimedOut` here?

executor/workflow_manager_sfn_test.go
```diff
@@ -725,6 +725,148 @@ func TestUpdateWorkflowStatusExecutionNotFoundStopRetry(t *testing.T) {
...
+	c.saveWorkflow(t, workflow)
+
+	sfnExecutionARN := c.manager.executionARN(workflow, c.workflowDefinition)
+	c.mockSFNAPI.EXPECT().
+		DescribeExecutionWithContext(gomock.Any(), &sfn.DescribeExecutionInput{
+			ExecutionArn: aws.String(sfnExecutionARN),
+		}).
+		Return(&sfn.DescribeExecutionOutput{
+			Status: aws.String(sfn.ExecutionStatusFailed), // when activity times out, execution immediately fails
+			Output: aws.String(`{ "error": "States.Timeout", "cause": null}`),
```
mohit: Am I correct to assume that `DescribeExecutionOutput.Output` is set for `ActivityTimedOut` but not for `ExecutionTimedOut`?

executor/workflow_manager_sfn_test.go
bgveenstra: I also want to check that `resources.StatusReasonJobTimedOut` contains the message we want. So I did: 1d70b57d8c74ae34ad02276208f3403f622ddda1. This test should probably be in the resource package, but it felt like a very minimal thing to add a file for.

executor/workflow_manager_sfn_test.go
bgveenstra: Nope, I just missed that line. Good catch! Removed 
 in 4c71abf85e1dee4cef533d86d85e55fac606d723

PR #96, opened 2017-11-16
Fix job start date to be when the activity was started.

kofi-clever:
Basically, don't set `Job.StartedAt` until the job actually starts, so we have an idea of time waiting in queue for historical jobs.

- [N/A] Run "make generate"

### DIFF
--- a/executor/workflow_manager_sfn.go
+++ b/executor/workflow_manager_sfn.go
@@ -465,6 +465,10 @@ func (wm *SFNWorkflowManager) UpdateWorkflowHistory(workflow *models.Workflow) e
 				job.ID = fmt.Sprintf("%d", aws.Int64Value(evt.Id))
 				job.Attempts = []*models.JobAttempt{}
 				job.CreatedAt = strfmt.DateTime(aws.TimeValue(evt.Timestamp))
+				if *evt.Type != sfn.HistoryEventTypeTaskStateEntered {
+					// Non-task states technically start immediately, since they don't wait on resources:
+					job.StartedAt = job.CreatedAt
+				}
 				job.Status = models.JobStatusCreated
 				if details := evt.StateEnteredEventDetails; details != nil {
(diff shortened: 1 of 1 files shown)


kofi-clever:
don't know if anything downstream relied on this always being set, but seemed like it should be safe

executor/workflow_manager_sfn.go
```diff
@@ -509,6 +507,7 @@ func (wm *SFNWorkflowManager) UpdateWorkflowHistory(workflow *models.Workflow) e
 				job.Status = models.JobStatusQueued
 			case sfn.HistoryEventTypeActivityStarted:
```
mohit: I'm not sure this state is exists for Choice and Success States. With this change we would never be setting job.StartedAt for them.

executor/workflow_manager_sfn.go
```diff
@@ -465,7 +465,6 @@ func (wm *SFNWorkflowManager) UpdateWorkflowHistory(workflow *models.Workflow) e
 				job.ID = fmt.Sprintf("%d", aws.Int64Value(evt.Id))
 				job.Attempts = []*models.JobAttempt{}
 				job.CreatedAt = strfmt.DateTime(aws.TimeValue(evt.Timestamp))
-				job.StartedAt = strfmt.DateTime(aws.TimeValue(evt.Timestamp))
```
mohit: seems like we want to set this for `HistoryEventType{Choice,Succeed}StateEntered` but not for `HistoryEventTypeTaskStateEntered`.

kofi-clever: yeah, good catch - will update to keep setting it for non-task states

kofi-clever: thanks for spotting that - https://github.com/Clever/workflow-manager/pull/96/commits/4aa55f11ffe570cc2d029043f65859b28bf7fd38

PR #100, opened 2017-11-30
Fix default queue

samfishman:
go-swagger wasn't doing the default for us (see https://github.com/Clever/workflow-manager/blob/18a48402b27c3b73804eb58d6edf4b29e7ff01b6/gen-go/models/start_workflow_request.go)

- [x] Run "make generate"

### DIFF
--- a/gen-go/models/start_workflow_request.go
+++ b/gen-go/models/start_workflow_request.go
@@ -24,7 +24,7 @@ type StartWorkflowRequest struct {
 	Namespace string `json:"namespace,omitempty"`
 
 	// queue
-	Queue *string `json:"queue,omitempty"`
+	Queue string `json:"queue,omitempty"`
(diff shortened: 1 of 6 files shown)


gen-go/models/start_workflow_request.go
```diff
@@ -24,7 +24,7 @@ type StartWorkflowRequest struct {
 	Namespace string `json:"namespace,omitempty"`
 
 	// queue
-	Queue *string `json:"queue,omitempty"`
+	Queue string `json:"queue,omitempty"`
```
rgarcia: `./handler.go:165:80: invalid indirect of req.Queue (type string)`

samfishman: I suppose I should compile code before submitting it

PR #101, opened 2017-12-01
[Infra-2652] add ResolvedByUser to workflow summary

bgveenstra:
Jira: [INFRA-2652](https://clever.atlassian.net/browse/INFRA-2652)

To support the use case of tracking which workflows have been investigated by a human, this PR adds a boolean `ResolvedByUser` field to the workflow model. Endpoint updates:

- a new `POST /workflows/{workflowID}/resolved` endpoint that sets the target workflow's `ResolvedByUser` to `true` if it is `false` but returns a `409 Conflict` if it's already `true`
- ability to send an optional `ResolvedByUser` boolean parameter to the `GET /workflows` endpoint that filters the results

**Open questions:**
- `ResolvedByUser` was intended to be used to mark that someone has checked on the reason for failure of a workflow.  Should we prevent users from "resolving" workflows that have certain (other) statuses?

**Other notes:**
- `ResolvedByUser` is an optional boolean parameter.  The `ResolvedByUserWrapper` allows us to distinguish between a `false` value was passed into a `GetWorkflows` request and a `false` default value that wasn't part of the request.  This is invisible to the client.
- The `GET /workflows` endpoint accepts a `status` parameter or a `resolvedByUser` parameter, but not at once. This should be clear in the documentation. 

**Pre-deploy checklist:**
- [x] Update swagger.yml version (to 0.7.1)
- [x] Run "make generate"


swagger.yml
```diff
@@ -167,12 +171,21 @@ paths:
...
+            The status of the workflow (queued, running, etc.). Cannot be sent in the
+             same request as the resolvedByUser parameter.
           in: query
           type: string
+        - name: resolvedByUser
+          in: query
+          type: boolean
+          description:
+            A flag that indicates whether the workflow has been marked resolved by a user. Cannot
+             be sent in the same request as the status parameter.
```
xavi-: Idea: Rename param to `resolved-only`.  That way we could avoid the need for `ResolvedByUserWrapper`.

handler_test.go
```diff
@@ -45,3 +45,55 @@ func TestValidateTagsMap(t *testing.T) {
...
+	definitionName := "defName"
+	// error if status and resolvedByUser are both sent
+	inputWithStatusAndResolvedTrue := &models.GetWorkflowsInput{
+		ResolvedByUser:         &boolTrue,
+		Status:                 &failedString,
+		WorkflowDefinitionName: definitionName,
+	}
+
+	workflowQuery, err := paramsToWorkflowsQuery(inputWithStatusAndResolvedTrue)
+	assert.Error(t, err)
```
xavi-: Might be worth also adding `assert.IsType(t, models.BadRequest{}, err)` check

store/dynamodb/dynamodb_store.go
```diff
@@ -582,15 +594,28 @@ func (d DynamoDB) GetWorkflowByID(id string) (models.Workflow, error) {
 
 // GetWorkflows returns all workflows matching the given query.
 func (d DynamoDB) GetWorkflows(query *models.WorkflowQuery) ([]models.Workflow, string, error) {
+	fmt.Println("DDB store getworkflows")
```
xavi-: Nit. Remove debugging print-statement.

swagger.yml
bgveenstra: And then we could have `unresolved-only` parameter to get just the unresolved.

swagger.yml
```diff
@@ -412,6 +450,8 @@ definitions:
       input:
         # format: json
         type: string
+      resolvedByUser:
+        type: boolean
```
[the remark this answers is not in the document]
bgveenstra: I don't think we should expose this in ark @xavi-. It's not really a property inherent in the workflow and might not always live in workflow-manager.

store/dynamodb/workflow.go
```diff
@@ -173,6 +181,80 @@ func (sk ddbWorkflowSecondaryKeyWorkflowDefinitionCreatedAt) KeySchema() []*dyna
 	}
 }
 
+// ===============================
```
bgveenstra: rm

swagger.yml
bgveenstra: Thinking through `GET /workflows` parameters:
- the original design, which had `resolvedByUser` to get unresolved or resolved
- this newer idea, which would be to use `unresolvedOnly` to get unresolved or all

New idea:
- removes `ResolvedByUserWrapper` complexity under the hood
- would lead to a `resolvedOnly` parameter if we want to get only resolved queries
- probably still uses the secondary key so we don't waste read capacity

I'm going to stick with `ResolvedByUser` because based on earlier discussion with @mohit, we don't particularly want to use two different params for this.

mohit:
The use case here for Hubble is for showing unresolved workflows. Does the
current design allow for getting only unresolved workflows?

bgveenstra:
@mohit Yes, it does.

PR #102, opened 2017-12-05
Add `Result` to `SLState`

nathanleiby:
This field is part of the states language spec, but we didn't yet support it.
I plan to use this to force a specific Result via a PASS state.

- [x] Run "make generate"


nathanleiby:
cc @mohit

swagger.yml
```diff
@@ -653,6 +653,8 @@ definitions:
         type: string
       ResultPath:
         type: string
+      Result:
```
nathanleiby: does this need `x-omitempty: true`?

rgarcia: nope, `x-omitempty` is only added to array fields where empty array `[]` is equivalent to omitting

rgarcia:
you'll have to pull in the client change in circle-ci-integrations, e.g. https://github.com/Clever/circle-ci-integrations/pull/29

PR #103, opened 2017-12-06
Add and set a CreatedAt date for previous job attempts.

kofi-clever:
More for completeness than anything.
We currently only have a start and stop date for failed job attempts. Defaulting created date to the created date of the latest job in hubble, but that looks off:

Adding created date as well for a more complete picture of how long a job took.

- [x] Run "make generate"


nathanleiby:
the code change looks good, but I'm afraid I'm not familiar with dates and why `createdAt` is giving you a weird value

kofi-clever:
yeah, that's more of a hubble bug - it currently copies the created date from the parent job object, which is set to the created date of the latest attempt

PR #104, opened 2017-12-06
Persist all previous job attempts when creating new ones.

kofi-clever:
Minor bug was causing the attempts array to contain only one previous attempt.

- [N/A] Run "make generate"


rgarcia:
nice catch!

PR #105, opened 2017-12-12
Set ResolvedByUser if workflow succeeded/cancelled

nathanleiby:
Previously, we only set this flag if a user manually submitted it. In
practice, we want a simple query for all "unresolved" workflows versus
"resolved" ones.

- Unresolved: queued, running, failed (but not yet investigated)
- Resolved: succeeded, cancelled, failed (and investigated)

This change auto-resolves succeeded and cancelled, and leaves it up to
the user to manually resolve failed workflows.

- [x] Run "make generate"


nathanleiby:
Removed the `swagger.yml` version bump, per our offline discussions. summary: Version bumps are only relevant to when the `swagger.yml` file itself changes, hence the client code is regenerated.

PR #106, opened 2017-12-12
Disallow cancelling failed workflows

nathanleiby:
https://clever.atlassian.net/browse/INFRA-2703

Previously, we allowed cancelling failed workflows as a workaround for
"resolving". We cancelled a failed job in order to manually acknowledge
that we were done investigating that job.

We recently added the concept ResolvedByUser in workflow-manager:
https://github.com/Clever/workflow-manager/pull/101

This means we can remove the overloaded idea of cancelling failed jobs,
and instead mark them as ResolvedByUser.

- [ ] Run "make generate"


nathanleiby:
This is blocked by deploying https://github.com/Clever/workflow-manager/pull/105, rolling out corresponding Hubble changes, and verifying the "cancel a failed workflow" behavior is no longer needed.

PR #107, opened 2017-12-27
Support wait states

nathanleiby:
Previously, we were not persisting the fields required to have `Wait` states. This change allows us to submit and persist those fields.

http://docs.aws.amazon.com/step-functions/latest/dg/amazon-states-language-wait-state.html

- [x] Run "make generate"


nathanleiby:
@mohit thoughts on how we can prevent adding these piecewise? 

Two ideas (non-exclusive):

- we could try to get `swagger.yml` fully updated to support all of states language (how much work would this be?)
- we could reject things submitted to WorkflowManager that have unsupported fields. => The reason I found this issue... I was posting objects that had the field `Seconds`, WorkflowManager was happily accepting them, but I was getting invalid schema errors upon submitting the workflow. I think it would have been less painful if WFM had rejected the workflow definition immediately.

bgveenstra:
I think we could have gotten better/earlier errors by waiting to add states to our list of `SLStateType`s until after they were actually supported.  At this point, the only one we don't support is `Parallel`:

  ```
SLStateType:
    type: string
    enum:
      - "Pass"
      - "Task"
      - "Choice"
      - "Wait"
      - "Succeed"
      - "Fail"
      - "Parallel"
```

We could remove `Parallel` from the list of allowed types until it's implemented.

PR #110, opened 2018-01-06
Set `ResolvedByUser = true` within `CancelWorkflow`

nathanleiby:
Followup to https://github.com/Clever/workflow-manager/pull/105

Previously, if a workflow was cancelled, it never got the resolvedByUser attribute set to true. We were attempting to update this bool during `UpdateWorkflowSummary`, but this function exits early if a `WorkflowIsDone` ... which includes being cancelled.

https://github.com/Clever/workflow-manager/blob/4fd44aeb0edeb7d46019cb5e355cdc83299d3533/executor/workflow_manager_sfn.go#L347

We now update it within `CancelWorkflow`

### DIFF
--- a/executor/workflow_manager_sfn.go
+++ b/executor/workflow_manager_sfn.go
@@ -303,6 +303,7 @@ func (wm *SFNWorkflowManager) CancelWorkflow(workflow *models.Workflow, reason s
 	}
 
 	workflow.StatusReason = reason
+	workflow.ResolvedByUser = true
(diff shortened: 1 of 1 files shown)


mohit:
I think the expected behavior here would be to *not* set `Workflow.Status = Cancelled` outside of the `UpdateWorkflowSummary` function. This would mean that we can continue to have the `resolvedByUser` flag only set in the `UpdateWorkflowSummary` function. This way we ensure that we don't mark Workflows as resolved that might fail to be updated as `cancelled`

PR #111, opened 2018-01-22
update go client after recent go-swagger update

bgveenstra:
- [x] Run "make generate"


gen-go/models/bad_request.go
```diff
@@ -32,21 +26,3 @@ func (m *BadRequest) Validate(formats strfmt.Registry) error {
 	}
 	return nil
 }
-
-// MarshalBinary interface implementation
```
bgveenstra: here

PR #114, opened 2018-02-02
Omitempty fix

rgarcia:
Ran `make generate` to pull in latest `wag` and regenerate with https://github.com/Clever/wag/pull/171

Updated mocks for new SFN methods


gen-go/models/s_l_state.go
```diff
@@ -17,13 +19,13 @@ import (
 type SLState struct {
 
 	// catch
-	Catch []*SLCatcher `json:"Catch"`
+	Catch []*SLCatcher `json:"Catch,omitempty"`
```
rgarcia: ☝️

PR #115, opened 2018-02-06
Update gomock and friends so workflow-manager plays nice with other repos.

bstein-clever:
cc @kvigen 

- [ ] Run "make generate"


mohit:
@bstein-clever please bump the swagger.yml version since this changes client code.

bstein-clever:
Bumped and squashed into the same commit.

PR #118, opened 2018-02-07
Bump version.

bstein-clever:
Let's bump it after all to make deps better. :sigh:


rgarcia:
... I'm confused. Why was this necessary?

bstein-clever:
I was getting an error locally when trying to update, because my gopkg.toml required v0.8.2, and master didn't qualify. Or so dep told me.

PR #124, opened 2018-03-09
Update Loop using SQS

nathanleiby:
JIRA: https://clever.atlassian.net/browse/INFRA-2829

This replaces polling dynamodb for updates (lots of calls against DynamoDB: looking up workflows, locking, and updating workflows) with a queue-based approach (puts workflows to update in a queue, no need for locking).

See mini-spec for more details: https://docs.google.com/document/d/1ghjxiImZYRN2ub233okATn2hG1McKhFm9yvm1rL_kPw/edit?ts=5a99edca#

I've tested locally and seems to work smoothly.

Considerations:

- I want to improve unit testing a little more
- Would it be worth abstracting the queue further? Lots of folks have made small abstractions around SQS that would help minimize the need to pass around config like delay duration, queue URL, etc (example: https://github.com/h2ik/go-sqs-poller)

--

- [x] Update swagger.yml version (n/a)
- [x] Run "make generate" (n/a)

### DIFF
--- a/executor/workflow_manager.go
+++ b/executor/workflow_manager.go
@@ -2,10 +2,10 @@ package executor
 
 import (
 	"context"
-	"fmt"
 	"time"
 
 	"github.com/Clever/workflow-manager/gen-go/models"
+	"github.com/Clever/workflow-manager/resources"
 	"github.com/Clever/workflow-manager/store"
 	"gopkg.in/Clever/kayvee-go.v6/logger"
 
@@ -44,7 +44,6 @@ func PollForPendingWorkflowsAndUpdateStore(ctx context.Context, wm WorkflowManag
 }
 
 func checkPendingWorkflows(ctx context.Context, wm WorkflowManager, thestore store.Store, sqsapi sqsiface.SQSAPI, sqsQueueURL string) (string, error) {
-	// TODO: use context
 	out, err := sqsapi.ReceiveMessageWithContext(ctx, &sqs.ReceiveMessageInput{
 		MaxNumberOfMessages: aws.Int64(1),
 		QueueUrl:            aws.String(sqsQueueURL),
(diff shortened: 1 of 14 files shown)


executor/workflow_manager.go
```diff
@@ -29,56 +35,60 @@ func PollForPendingWorkflowsAndUpdateStore(ctx context.Context, wm WorkflowManag
 			ticker.Stop()
 			return
 		case <-ticker.C:
-			if id, err := checkPendingWorkflows(wm, thestore); err != nil {
-				log.ErrorD("poll-for-pending-workflows", logger.M{"id": id, "error": err.Error()})
-			} else {
-				log.InfoD("poll-for-pending-workflows", logger.M{"id": id})
+			out, err := sqsapi.ReceiveMessageWithContext(ctx, &sqs.ReceiveMessageInput{
```
rgarcia: how should we think about the 500 ms ticker around this call to `ReceiveMessage` now? Is there a limit on calls to ReceiveMessage?

executor/workflow_manager.go
```diff
@@ -29,56 +35,60 @@ func PollForPendingWorkflowsAndUpdateStore(ctx context.Context, wm WorkflowManag
...
-		if err := thestore.UnlockWorkflow(wfLockedID); err != nil {
-			log.ErrorD("pending-workflows-unlock-error", logger.M{"id": wfLockedID, "error": err.Error()})
-		}
-	}()
+	// If workflow is not yet complete, send message to SQS to request a future update.
+	if !resources.WorkflowIsDone(&wf) {
+		_, err = sqsapi.SendMessageWithContext(ctx, &sqs.SendMessageInput{
+			MessageBody:  aws.String(wfID),
+			QueueUrl:     aws.String(sqsQueueURL),
+			DelaySeconds: aws.Int64(30),
```
rgarcia: consider defining as constant and documenting with comment

store/store.go
```diff
@@ -25,14 +24,8 @@ type Store interface {
 	UpdateWorkflow(workflow models.Workflow) error
 	GetWorkflowByID(id string) (models.Workflow, error)
 	GetWorkflows(query *models.WorkflowQuery) ([]models.Workflow, string, error)
-	GetPendingWorkflowIDs() ([]string, error)
-	LockWorkflow(id string) error
-	UnlockWorkflow(id string) error
```
rgarcia: 🎉

rgarcia:
re: abstracting queue further, I'm fine not doing that, since the ops here are pretty straightforward and everyone is pretty familiary with aws-sdk-go.

executor/workflow_manager.go
nathanleiby: Added a constant with comment https://github.com/Clever/workflow-manager/pull/124/commits/f9e5569efb7272d0477c58f7d53330de4c545317

executor/workflow_manager.go
nathanleiby: I don't see a specific limit. Seems like the limits are really high. Other examples don't include a ticker. 

I'm inclined to remove it.

https://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/sqs-limits.html

nathanleiby: removed ticker f79145c

PR #125, opened 2018-03-12
don't save Jobs for very large workflows

mohit:
- DynamoDB does not like items > 400KB. 

Therefore we can return a large workflow but not save it

Additional things we could do here:

- limit by number of jobs. This would mean we don't have to retry
- exit out of UpdateWorkflowDetail early for large workflows. We should have some kind of limit to prevent against Hubble reloading a large workflow too often, since it hits our AWS limits hard. 
- binary encode jobs array so that we can save jobs in ddb (I don't see a strong reason for this right now)

### DIFF
--- a/store/dynamodb/dynamodb_store.go
+++ b/store/dynamodb/dynamodb_store.go
@@ -19,6 +19,9 @@ import (
 	"gopkg.in/Clever/kayvee-go.v6/logger"
 )
 
+// AWS ValidationException message when item is > 400KB
+const errMessageItemTooLarge = "Item size has exceeded the maximum allowed size"
+
 type DynamoDB struct {
 	ddb         dynamodbiface.DynamoDBAPI
 	tableConfig TableConfig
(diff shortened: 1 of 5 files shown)


store/dynamodb/dynamodb_store.go
```diff
@@ -530,8 +533,22 @@ func (d DynamoDB) UpdateWorkflow(workflow models.Workflow) error {
...
 				return store.NewNotFound(workflow.ID)
+			case "ValidationException":
+				if awsErr.Message() == errMessageItemTooLarge {
+					log.WarnD("workflow-too-large", logger.M{
+						"error":     awsErr.Message(),
+						"id":        workflow.ID,
+						"name":      workflow.WorkflowDefinition.Name,
+						"namespace": workflow.Namespace,
+					})
+					// try again without jobs
```
bgveenstra: Code LGTM, but please document this behavior in the README before merging.

mohit: https://github.com/Clever/workflow-manager/pull/125/commits/846f3413bfb3c30da82caae000216234a820090f

PR #127, opened 2018-03-15
Backoff in polling loop DynamoDB throughput error

nathanleiby:
In the case that many jobs are submitting at once, the SQS approach can lead to increased load on DynamoDB.

To help mitigate this, let's backoff in our polling loop if we start to see DynamoDB throughput errors.

--

This change also removes references to a now unused dynamodb GSI 14c2bc0

### DIFF
--- a/executor/workflow_manager.go
+++ b/executor/workflow_manager.go
@@ -3,6 +3,7 @@ package executor
 import (
 	"context"
 	"sync"
+	"time"
 
 	"github.com/Clever/workflow-manager/gen-go/models"
 	"github.com/Clever/workflow-manager/resources"
@@ -10,6 +11,8 @@ import (
 	"gopkg.in/Clever/kayvee-go.v6/logger"
 
 	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/awserr"
+	"github.com/aws/aws-sdk-go/service/dynamodb"
 	"github.com/aws/aws-sdk-go/service/sqs"
 	"github.com/aws/aws-sdk-go/service/sqs/sqsiface"
 )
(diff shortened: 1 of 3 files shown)


executor/workflow_manager.go
```diff
@@ -23,10 +26,20 @@ type WorkflowManager interface {
...
+var backoffDuration = time.Second * 5
+var shouldBackoff = false
+var shouldBackoffOnce *sync.Once
+
 // PollForPendingWorkflowsAndUpdateStore polls an SQS queue for workflows needing an update.
 // It will stop polling when the context is done.
 func PollForPendingWorkflowsAndUpdateStore(ctx context.Context, wm WorkflowManager, thestore store.Store, sqsapi sqsiface.SQSAPI, sqsQueueURL string) {
 	for {
+		if shouldBackoff {
+			time.Sleep(backoffDuration)
```
xavi-: Nit: maybe emit a log when in `backoff` mode.
A More minor nit:  Do we want the backoff to increase after repeated thottles?

nathanleiby: added log for polling backoff 472650c

decided to not to exponential backoff for now -- let's see how this behaves and improve as needed

executor/workflow_manager.go
```diff
@@ -23,10 +26,21 @@ type WorkflowManager interface {
...
 
+var backoffDuration = time.Second * 5
+var shouldBackoff = false
+var shouldBackoffOnce *sync.Once
+
 // PollForPendingWorkflowsAndUpdateStore polls an SQS queue for workflows needing an update.
 // It will stop polling when the context is done.
 func PollForPendingWorkflowsAndUpdateStore(ctx context.Context, wm WorkflowManager, thestore store.Store, sqsapi sqsiface.SQSAPI, sqsQueueURL string) {
 	for {
+		if shouldBackoff {
```
mohit: nit: can we pull out the `backoff` into a struct and then use that in the function? Or could we add backoff state to the context?

nathanleiby: I agree it could be more elegant, but I'd prefer to stick with a bare-minimum approach for now. 

If we find it needs to be more robust/fancy, I'm happy to improve later.

PR #129, opened 2018-03-16
simplify backoff; remove concurrrency from poll/update loop

nathanleiby:
The concurrent approach allowed us to do many updates in parallel, but
we don't need this performance optimization. Instead, our priority is
to prevent the polling loop from causing too many DB writes, leading to
throttling.

Previously, 10 workflows were concurrently being updated, and each of
those updates was internally doing retries via aws-sdk-go library.  With
this change, we'll now backoff immediately the first time the polling
loop sees a throttled error. Ideally, this should result in up to 10x less
throttled writes.

More on AWS SDK Go retry configuration:
https://github.com/aws/aws-sdk-go/blob/952498f4a390118ac65ad59cbe0c8b57ed69b6b5/aws/client/client.go#L58-L64

### DIFF
--- a/executor/workflow_manager.go
+++ b/executor/workflow_manager.go
@@ -2,7 +2,6 @@ package executor
 
 import (
 	"context"
-	"sync"
 	"time"
 
 	"github.com/Clever/workflow-manager/gen-go/models"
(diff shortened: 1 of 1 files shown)


rgarcia:
if in the future we want both concurrency and behavior where the loop stops at the first throttling error, we could

1. create a new `context.WithCancel`
2. pass this context to concurrent calls to `updatePendingWorkflow`
3. if we see a throttling error, cancel the context

nathanleiby:
Sounds good! I considered that but felt in the short term that this was
simpler to work with.

Thanks for laying it out clearly and we can revisit if needed.

PR #130, opened 2018-03-16
don't pull workflow history for more than a minute

mohit:
This will make sure we don't just keep polling while the ALB has timed out. 

Also going to add a task for next sprint to support handler level timeout context


mohit:
cc: @meliaj just FYI

executor/workflow_manager_sfn_test.go
```diff
@@ -847,6 +847,7 @@ func TestUpdateWorkflowStatusWorkflowTimedOut(t *testing.T) {
 	require.NoError(t, c.manager.UpdateWorkflowSummary(workflow))
 	require.NoError(t, c.manager.UpdateWorkflowHistory(workflow))
 	assert.Equal(t, models.WorkflowStatusFailed, workflow.Status)
+	// TODO: this line below is not testing anything. Verify what needs to go in this assertion
```
bgveenstra: I'm not sure what the future goal for this line is -- pls just check that the comment is still useful for you

mohit: it's for me to get back to this next week since the assert.Equal is a noop. I want to make it not a noop :).

PR #131, opened 2018-03-20
add server-level timeout

rgarcia:
- use `http.TimeoutHandler` to send a 5XX response after a 5s timeout
- use custom middleware to cancel the `http.Request`'s context object
- pass request context to AWS API calls

end effect should be:
- clients of workflow manager will receive a 5xx after 5s
- the workflow-manager goroutine for a timed-out request will stop making AWS requests

### DIFF
--- a/executor/workflow_manager_sfn_test.go
+++ b/executor/workflow_manager_sfn_test.go
@@ -141,6 +141,8 @@ func TestCreateWorkflow(t *testing.T) {
 	input := "{\"json\": true}"
 
 	t.Run("CreateWorkflow for existing StateMachines", func(t *testing.T) {
+		ctx, cancel := context.WithCancel(context.Background())
+		defer cancel()
 		c := newSFNManagerTestController(t)
 		defer c.tearDown()
 		stateMachineArn := stateMachineARN(c.manager.region, c.manager.accountID,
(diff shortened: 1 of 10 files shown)


main.go
```diff
@@ -75,7 +76,20 @@ func main() {
...
+		},
+		func(handler http.Handler) http.Handler {
+			return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+				newCtx, cancel := context.WithTimeout(r.Context(), timeout)
+				defer cancel()
+				r = r.WithContext(newCtx)
+				handler.ServeHTTP(w, r)
+			})
+		},
+	})
```
rgarcia: cc @natebrennand

executor/workflow_manager_sfn_test.go
```diff
@@ -141,6 +141,8 @@ func TestCreateWorkflow(t *testing.T) {
 	input := "{\"json\": true}"
 
 	t.Run("CreateWorkflow for existing StateMachines", func(t *testing.T) {
+		ctx, cancel := context.WithCancel(context.Background())
```
natebrennand: why can't we just use `context.Background()` here?

rgarcia: there are parts of w-m that spin up background threads / polling loops, so using context.Background() (which never is canceled) might be leaky compared a context that we cancel at the end of the test

PR #134, opened 2018-04-03
Remove `queued` workflow status

nathanleiby:
Currently, a "queued" workflow means something more like "initialized". We have created a Workflow but have not yet sync'd its status from Step Functions.

There's not a huge amount of value in having a separate state for this. Some users have found it confusing.  In the short-run, might we just removed "queued"?

There's one specific loss in resolution here. Sometimes "queued" helps us catch wedged workflows. It's possible for us to fail to create an execution in Step Functions and also fail to cleanup the workflow. Having it in a "queued" state vs "running" helps us notice that this workflow is broken.

- [x] Run "make generate"

--

In the long-run, we might prefer adding more Workflow Statuses to give us more insight.

1. "initialized" -- we've submitted the workflow to Step Functions, but haven't yet sync'd any info back to Workflow Manager
2. "queued" - the Step Function is "running", but its first task is "queued"... Basically, there are no workers yet processing the workflow. From a user's perspective, this means that queue size == backlog ... It may indicate that more workers need to be run in order to catch up.
3. "running" - the Step Function is "running" and its first task is "running" or "completed".

This approach would require us to fetch the first page of tasks from each workflow, so would require more API calls to Step Functions.

### DIFF
--- a/docs/definitions.md
+++ b/docs/definitions.md
@@ -312,7 +312,7 @@
 
 <a name="workflowstatus"></a>
 ### WorkflowStatus
-*Type* : enum (queued, running, failed, succeeded, cancelled)
+*Type* : enum (running, failed, succeeded, cancelled)
 
 
 <a name="workflowsummary"></a>
--- a/docs/overview.md
+++ b/docs/overview.md
@@ -17,3 +17,6 @@ Orchestrator for AWS Step Functions
 ### Produces
 
 * `application/json`
+
+
+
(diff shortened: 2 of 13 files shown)


nathanleiby:
from slack

person1: 
> personally i think of “queued” meaning *any* task in the workflow is queued (not just the first)

person2:
> I agree. I would be fine with the separation of "none have started" and "waiting on one of the ones in the middle", but to me "running" should mean "data is actively being processed".

nathanleiby:
Closing as wontfix.

PR #135, opened 2018-04-05
remove lock table

rgarcia:
no longer used in new update loop


mohit:
also remove from Infra?

PR #136, opened 2018-04-06
adding update loop for this workflow

mohit:
cc: @nathanleiby


nathanleiby:
Makes sense to me. 

Would it make sense to pull any of this into a help that Create and Retry share, to ensure they stay in sync in the future? (rule of 3 tho, so maybe not yet)

nathanleiby:
Going to merge for now for stability. Can revisit later if we want to refactor.

PR #143, opened 2018-08-02
[INFRA-3134] workflow-manager is wag so needs node

bgveenstra:
Since `workflow-manager` is a WAG service, this PR:
- uses the version of the base docker image with node (by SHA, to lock in exact versions)
- unrelated, remvoes some uneccessary line breaks

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -3,7 +3,7 @@ jobs:
   build:
     working_directory: /go/src/github.com/Clever/workflow-manager
     docker:
-    - image: circleci/golang:1.10-node
+    - image: circleci/golang@sha256:dbcd25bb717ed97b229311388efd6a1a3e1841da08899f40b34632ebb58915ed # circleci/golang:1.10-node
     steps:
     - checkout
     - setup_remote_docker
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -3,7 +3,7 @@ jobs:
   build:
     working_directory: /go/src/github.com/Clever/workflow-manager
     docker:
-    - image: circleci/golang:1.10
+    - image: circleci/golang:1.10-node
```
natebrennand: assuming this is just Go 1.10 + node?
Should this image come with npm installed? It might be nice to have the image tag note which node version?

bgveenstra: This is a CircleCI-provided "convenience image." It includes `node 8.11.3` (LTS) and `npm 5.6.0`.  

That's not the latest `npm`, but we probably don't need to over-optimize by installing the newest one.  Removed steps to update `npm` in 8a8b39b. 

For wag repos, I haven't found any place where we specifically coded a node version to use. So, I added a sha to the circleci convenience image to lock it in at 8.11.3, in 647fcc2.

PR #147, opened 2018-10-04
Fix 404 crash

xavi-:
Currently, going to this URL will crash workflow-manager: https://production--workflow-manager.int.clever.com/workflow-definitions/alskdjf/0

This PR is trying to fix that bug

- Upgraded to the latest WAG
- Removed references to WAG DB code.  The latest WAG had backwards incompatible changes to the DB package.  The only function that used this code was `GetWorkflowDefinition`.  It now uses the dynamo library directly.
- Changed some log levels to Trace

cc @mohit

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -2,7 +2,6 @@
 
 
 [[projects]]
-  branch = "master"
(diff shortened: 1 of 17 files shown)


mohit:
LGTM

PR #148, opened 2018-10-04
The /workflows endpoint can now filter by both status and resolvedByUser

xavi-:
Workflow-manager now allows people filter to filter by `status` is `resolvedByUser`.  It uses a [Filter Query](https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/Query.html#Query.FilterExpression) to ensure the query remain performant.  Filter Queries filter after they query is completed.  In other words:
> a Query will consume the same amount of read capacity, regardless of whether a filter expression is present.

cc @mohit

### DIFF
--- a/store/dynamodb/dynamodb_store.go
+++ b/store/dynamodb/dynamodb_store.go
@@ -585,7 +585,6 @@ func (d DynamoDB) GetWorkflows(ctx context.Context, query *models.WorkflowQuery)
 	// Use resolvedByUser index if querying by both Status and isReolvedByUser.  The resolvedByUser
 	// index is smaller and typically shrinks over time, so it should be faster to query
 	if resolvedByUserIsSet {
-		// otherwise, if query includes a ResolvedByUser value that is set, query with the ResolvedByUser value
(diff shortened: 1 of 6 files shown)


store/dynamodb/dynamodb_store.go
```diff
@@ -581,17 +581,26 @@ func (d DynamoDB) GetWorkflows(ctx context.Context, query *models.WorkflowQuery)
...
-	}
 
-	if statusIsSet {
-		// if query includes status, query by status
-		dbQuery, err = ddbWorkflowSecondaryKeyDefinitionStatusCreatedAt{}.ConstructQuery(query)
-	} else if resolvedByUserIsSet {
+	// Use resolvedByUser index if querying by both Status and isReolvedByUser.  The resolvedByUser
+	// index is smaller and typically shrinks over time, so it should be faster to query
+	if resolvedByUserIsSet {
 		// otherwise, if query includes a ResolvedByUser value that is set, query with the ResolvedByUser value
```
bgveenstra: This comment no longer makes sense in new context.

xavi-: Whoops.  Removed in 3aef4fa40c7da856a4ab545b34654d2b4f2fd014

PR #150, opened 2018-10-19
Send notification to #oncall-infra if an unknown workflow is started

xavi-:
If jobs are being dropped on the floor, we should know about it

Jira: https://clever.atlassian.net/browse/INFRA-3274

### DIFF
--- a/handler.go
+++ b/handler.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 
 	"github.com/aws/aws-sdk-go/aws"
+	"gopkg.in/Clever/kayvee-go.v6/logger"
 
 	"github.com/Clever/workflow-manager/executor"
 	"github.com/Clever/workflow-manager/gen-go/models"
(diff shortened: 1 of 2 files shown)


handler.go
```diff
@@ -5,13 +5,16 @@ import (
...
 	"github.com/aws/aws-sdk-go/aws"
+	"gopkg.in/Clever/kayvee-go.v6/logger"
 
 	"github.com/Clever/workflow-manager/executor"
 	"github.com/Clever/workflow-manager/gen-go/models"
 	"github.com/Clever/workflow-manager/resources"
 	"github.com/Clever/workflow-manager/store"
 )
 
+var kvlog = logger.New("workflow-manager")
```
natebrennand: Should we just use the logger from the kayvee middleware?

handler.go
```diff
@@ -149,7 +152,15 @@ func (h Handler) StartWorkflow(ctx context.Context, req *models.StartWorkflowReq
...
 	}
-	if err != nil {
+	switch err.(type) {
+	case nil: // Do nothing
+	case models.NotFound:
+		kvlog.WarnD("start-unknown-workflow", logger.M{
+			"name":    req.WorkflowDefinition.Name,
+			"version": req.WorkflowDefinition.Version,
+		})
+		return &models.Workflow{}, err
```
natebrennand: fallthrough? 🤷‍♂️ 
```suggestion
		fallthrough
```

xavi-: I tried that originally.  Unfortunately you can't use `fallthrough`'s when switching on types.

handler.go
xavi-: I considered that but the middleware doesn't have the workflow's name

handler.go
```diff
@@ -149,7 +152,15 @@ func (h Handler) StartWorkflow(ctx context.Context, req *models.StartWorkflowReq
 	} else {
 		workflowDefinition, err = h.store.GetWorkflowDefinition(ctx, req.WorkflowDefinition.Name, int(req.WorkflowDefinition.Version))
 	}
-	if err != nil {
+	switch err.(type) {
+	case nil: // Do nothing
+	case models.NotFound:
+		kvlog.WarnD("start-unknown-workflow", logger.M{
```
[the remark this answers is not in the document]
xavi-: Force pushed change in e7223b4f0163e8b254f2bdd6792b41c0d1571677

PR #155, opened 2018-10-30
Reducing backoff time, moving it inside the loop, and checking for fa…

toddobryan-clever:
…iled write to SQS queue

- [ ] Run "make generate"

### DIFF
--- a/executor/workflow_manager.go
+++ b/executor/workflow_manager.go
@@ -25,18 +25,18 @@ type WorkflowManager interface {
 	UpdateWorkflowHistory(ctx context.Context, workflow *models.Workflow) error
 }
 
-var backoffDuration = time.Second * 5
+var backoffDuration = time.Second * 1
 var shouldBackoff = false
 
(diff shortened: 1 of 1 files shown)


executor/workflow_manager.go
```diff
@@ -25,18 +25,18 @@ type WorkflowManager interface {
...
-		if shouldBackoff {
-			log.WarnD("poll-for-pending-workflows-backoff", logger.M{"duration": backoffDuration.String()})
-			time.Sleep(backoffDuration)
-		}
-		shouldBackoff = false
+		//if shouldBackoff {
+		//	log.WarnD("poll-for-pending-workflows-backoff", logger.M{"duration": backoffDuration.String()})
+		//	time.Sleep(backoffDuration)
+		//}
+		//shouldBackoff = false
```
xavi-: Nit. If it were me, I would delete this code and the `shouldBackoff` variable

PR #159, opened 2018-10-31
update loop should not look at jobs array when deciding to requeue

rgarcia:
I took a peek at the SQS queue and found some workflows that were
really old, but still in the queue. Turns out these workflows were
finished, but had jobs that were still "in progress". This would
happen if someone loaded it in hubble (triggering an update of the
jobs array), but then never again loaded it.

The workflow.Jobs array is kind of a dumpster fire right now, so I'm
not entirely surprised by this bug... in the long run I think it
should be removed from the workflow object.

Here's an example workflow with the problem:

id: 53332426-43c8-4746-bcdb-47f8f3f545a8

discovered it by polling the SQS queue in the console and noticing that it was showing up more than once

looked it up in dynamo and saw that it was started / finished two weeks ago:
image
image

Note the final job in the jobs array shows up as "running"
Discovered that the update loop is looking at the jobs array, despite the jobs array no longer getting updated in the update loop. This PR fixes that.

### DIFF
--- a/executor/workflow_manager.go
+++ b/executor/workflow_manager.go
@@ -119,7 +119,7 @@ func updatePendingWorkflow(ctx context.Context, m *sqs.Message, wm WorkflowManag
 	// and re-queue a new message if the worfklow remains pending.
 	defer func() {
 		deleteMsg()
-		if !resources.WorkflowIsDone(&wf) {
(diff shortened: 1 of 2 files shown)


rgarcia:
cc @xavi- @toddobryan-clever another bug I just discovered, going to merge now since the queue is growing again

PR #160, opened 2018-11-26
remove creation of VERSION file

natebrennand:


natebrennand:
I believe I broke this with https://github.com/Clever/ci-scripts/pull/49

natebrennand:
We reverted the ci-scripts change because too many repos were following this pattern to make it worthwhile.

PR #161, opened 2018-12-04
Updated process-metrics to fix metrics bug

xavi-:
Process-metrics had a type-o ([`guage`](https://production--haproxy-logs.int.clever.com/_plugin/kibana/#/visualize/create?indexPattern=%5Blogs-%5DYYYY-MM-DD&type=histogram&_a=(filters:!(),linked:!f,query:(query_string:(analyze_wildcard:!t,lowercase_expanded_terms:!f,query:'type:guage')),vis:(aggs:!((id:'3',params:(field:container_app,orderBy:'2',size:20),schema:segment,type:terms),(id:'2',schema:metric,type:count)),type:histogram))&_g=(refreshInterval:(display:Off,pause:!f,section:0,value:0),time:(from:now-1w,mode:relative,to:now))) vs `gauge`) which causes errors in kinesis-alerts-consumer.

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -15,8 +15,8 @@
 [[projects]]
   name = "github.com/Clever/go-process-metrics"
   packages = ["metrics"]
-  revision = "86b8beeca7d3dc0ce41b3d29debfc47819eeb4b1"
-  version = "0.1.0"
+  revision = "ca5b011e155bf113777359fef3118d3122fffb54"
+  version = "0.2.0"
 
 [[projects]]
   name = "github.com/PuerkitoBio/purell"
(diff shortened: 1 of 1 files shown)


Gopkg.lock
```diff
@@ -428,12 +428,12 @@
 [[projects]]
   name = "gopkg.in/yaml.v2"
```
toddobryan-clever: Are we standardized on `ghodss/yaml` or do we use different YAML libraries depending on the repo?

xavi-: I'm not 100% sure to be to be honest.  My guess is that it's brought in by a dependency.

PR #164, opened 2019-02-26
embedded workflow-manager

rgarcia:
https://clever.atlassian.net/browse/INFRA-3373


embedded/poll.go
```diff
@@ -0,0 +1,208 @@
...
+
+var log = logger.New("wfm-embedded")
+
+// PollForWork begins polling for work. It stops when the context is canceled
+// or an un-recoverable error is encountered.
+func (e Embedded) PollForWork(ctx context.Context) error {
+	// register activities with AWS and spawn GetActivityTask polling loops
+	g, ctx := errgroup.WithContext(ctx)
+	for resourceName, resource := range e.resources {
+		activityArn := sfnconventions.EmbeddedResourceArn(resourceName, e.sfnRegion, e.sfnAccountID, e.environment)
```
mohit: If I understand the current implementation it seems that work submitted by one process can be picked up by another. i.e. If 2 containers (x, y) of app A are running then the work submitted by x might be picked up in y. 

It seems like that might cause confusion. I wonder if we can use container id or os.hostname() to protect against that?

executor/sfnconventions/sfnconventions.go
```diff
@@ -0,0 +1,70 @@
+package sfnconventions
```
mohit: 💯

embedded/poll.go
rgarcia: it's by design so that executions are robust against container crashes. Activities are still scoped to the environment they run in (and now that I look at this, they should probably be scoped by the app name as well). In general it feels like a slippery slope to let containers hold state between steps.

embedded/README.md
```diff
@@ -0,0 +1,38 @@
+# Embedded workflow-manager
+
+Embedded workflow manager implements a self-contained version of workflow-manager.
+It allows applications to create a set of workflow definitions specific that are isolated to the application itself.
```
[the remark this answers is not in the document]
rgarcia:
```suggestion
It allows applications to create a set of workflow definitions that are isolated to the application itself.
```

executor/sfnconventions/sfnconventions.go
rgarcia: it felt good to organize these methods ha

embedded/embedded.go
```diff
@@ -0,0 +1,411 @@
...
+		sfnAPI:              config.SFNAPI,
+		resources:           r,
+		workflowDefinitions: wfdefs,
+		workerName:          wn,
+	}, nil
+}
+
+const lettersAndNumbers = "abcdefghijklmnopqrstuvwxyz0123456789"
+
+func randString(n int) string {
```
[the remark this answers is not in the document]
rgarcia: I copied this from somewhere on the internet so I have no justification as to why :)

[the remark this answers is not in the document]
rgarcia: I guess the one justification is that it's nice to have control over the characters that go into the string

embedded/poll.go
rgarcia: a36ab42

PR #165, opened 2019-03-28
embedded: add support for NewWorkflowDefinition

rgarcia:
this will let programs create workflow definitions at runtime, in
addition to the wfdefs loaded on startup

### DIFF
--- a/embedded/embedded.go
+++ b/embedded/embedded.go
@@ -161,15 +161,15 @@ func verifyWorkflowDefinitionResources(wfdefs []models.WorkflowDefinition, resou
 	return nil
 }
 
-func (e Embedded) HealthCheck(ctx context.Context) error {
+func (e *Embedded) HealthCheck(ctx context.Context) error {
 	return nil
 }
 
-func (e Embedded) GetWorkflowDefinitions(ctx context.Context) ([]models.WorkflowDefinition, error) {
+func (e *Embedded) GetWorkflowDefinitions(ctx context.Context) ([]models.WorkflowDefinition, error) {
 	return e.workflowDefinitions, nil
 }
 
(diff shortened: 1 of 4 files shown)


embedded/poll.go
```diff
@@ -19,12 +19,12 @@ import (
 	"golang.org/x/time/rate"
 	"gopkg.in/Clever/kayvee-go.v6/logger"
 )
-
+n
```
Sayan-: ✂️

embedded/embedded.go
```diff
@@ -391,6 +391,26 @@ func (e Embedded) GetWorkflowByID(ctx context.Context, workflowID string) (*mode
 	}, nil
 }
 
+// NewWorkflowDefinition creates a new workflow definition.
```
mohit: maybe-an-aside: since at this point we would know the environment and version for the workflow definition can we just post it to sfn to be able to get a early error if the workflow-definition is incorrect?

embedded/embedded_test.go
```diff
@@ -0,0 +1,52 @@
...
+func (n newWorkflowDefinitionTest) Run(t *testing.T) {
+	ctx := context.Background()
+	_, err := n.wfm.NewWorkflowDefinition(ctx, n.input)
+	require.Equal(t, n.expected, err)
+	if n.assertions != nil {
+		n.assertions(ctx, t, n.wfm)
+	}
+}
+
+var newWorkflowDefinitionTests = []newWorkflowDefinitionTest{
```
Sayan-: might be nice to add a test case for the `workflow definition already exists` error

Sayan-:
Another aside - how do we version this embedded wfm? We don't currently widely use it, but it feels like we should probably consider versioning soon

Sayan-:
We additionally define a method `ParseWorkflowDefinition` that will unmarshal a single workflow definition from a yaml string.

PR #172, opened 2019-04-23
Checks to make sure cancelled workflows are marked resolvedByUser

toddobryan-clever:
- [ ] Run "make generate"


toddobryan-clever:
The one thing I'm not sure about is whether I need to manually add the cancelled workflow onto the update queue when it is cancelled.

mohit:
Can we add a test case for this?

toddobryan-clever:
Yeah. Was trying to think how I'd do that...

PR #173, opened 2019-04-23
update gomock dependency version

johnhuangclever:
- [ ] Run "make generate"


natebrennand:
ooh, looks like we're getting wag 3.0 here

johnhuangclever:
will roll out clever-dev->canary->prod

PR #175, opened 2019-04-25
use ES for GetWorkflows

rgarcia:
Hubble now directly hits ES instead of this endpoint. The remaining consumers of this endpoint are `ark` and `workflow-auto-resolver`, on the order of <10 requests a day. We currently provision two massive dynamo secondary indexes to power these queries + have written a ton of code to perform those queries. This PR switches the endpoint to hit elasticsearch instead of using dynamo.


handler.go
```diff
@@ -181,52 +190,133 @@ func (h Handler) StartWorkflow(ctx context.Context, req *models.StartWorkflowReq
...
+		var workflow models.Workflow
+		if err := json.Unmarshal(workflowBs, &workflow); err != nil {
+			return nil, err
+		}
+		workflows = append(workflows, workflow)
+	}
+	return workflows, nil
+}
+
 // GetWorkflows returns a summary of all workflows matching the given query.
```
bgveenstra: Could you update this comment to mention pagination?

rgarcia:
I should have mentioned: I opted for manual testing for this logic since it gets hit very infrequently, and I think we can actually pare down the interface exposed here even further if we took a look at the actual queries that ark and the lambda make.

handler.go
```diff
@@ -181,52 +190,133 @@ func (h Handler) StartWorkflow(ctx context.Context, req *models.StartWorkflowReq
...
+			"Workflow.status",
+			"Workflow.namespace",
+			"Workflow.queue",
+			"Workflow.input",
+			"Workflow.resolvedByUser",
+			"Workflow.retryFor",
+			"Workflow.retries",
+		),
+	}
+	req = append(req, h.es.Search.WithBody(strings.NewReader(h.getWorkflowsInputToESQuery(input))))
```
natebrennand: why is this appended and not just inlined on the slice declaration?

prime-time:
🥇

handler.go
rgarcia: that is... a good question haha. I think it's a relic of how i initially organized this code. Will fix

natebrennand: cool, doesn't seem like there's be any effect, was just confused

PR #171, opened 2019-04-23
workflow last job info

prime-time:
The hubble/repairs page is slow to load when there are a large number of active workflows. The reason for the slowness is that it iterates through every active workflow and loads the full workflow history for each one. We are in the process of modifying Hubble to query Elasticsearch for active and recent workflows instead of using workflow-manager, and we are also exploring the possibility of querying Elasticsearch for failed workflows. This change introduces a new method that fetches the most recent events in the state machine's execution history, and stores the last Job info in the `WorkflowSummary`. The new method shares a lot of logic with https://github.com/Clever/workflow-manager/blob/f0f8abee201526cbe7f51ee93fcb41dbf6d0cd27/executor/workflow_manager_sfn.go#L372, and the main difference that it doesn't page through the entire execution history to fill out the Jobs array and instead just looks at the most recent events. Some of the code from `UpdateWorkflowHistory` is refactored into separate functions so that they can be used in the new `updateWorkflowLastJob` method. 

Here's an earlier draft at this that would fill in the entire jobs array https://github.com/Clever/workflow-manager/pull/170. The downside of that approach is that it would call `GetExecutionHistory` for each failed workflow update, and that endpoint has a relatively low rate limit.

Below is an example workflow that I created with `lastJob` populated
```
[renatoprime@Renatos-MacBook-Pro workflow-manager (SYNC-1172-workflow-last-attempted-state-v2)]$ curl https://vpc-workflows-dev-2nfzpvqce62oumm6zcnjaipife.us-west-1.es.amazonaws.com/clever-dev-workflow-manager-dev-v3-workflows/default/fa275130-4be8-42be-85ac-7a6cdf378268 | jq
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100  2055  100  2055    0     0  17471      0 --:--:-- --:--:-- --:--:-- 17415
{
  "_index": "clever-dev-workflow-manager-dev-v3-workflows",
  "_type": "default",
  "_id": "fa275130-4be8-42be-85ac-7a6cdf378268",
  "_version": 3,
  "found": true,
  "_source": {
    "Workflow": {
      "createdAt": "2019-04-23T15:00:53.795429Z",
      "id": "fa275130-4be8-42be-85ac-7a6cdf378268",
      "input": "{\"uh\":\"boo\"}",
      "lastJob": {
        "createdAt": "2019-04-23T15:00:54Z",
        "id": "2",
[...]

PR #179, opened 2019-05-07
fix event ID to jobs map scoping

prime-time:
This change fixes a bug introduced when we did some refactoring here https://github.com/Clever/workflow-manager/pull/171/commits/ad8e0be85ca39a25acd9189a74a197f576261120. The map `eventIDToJob` was declared outside of a function and we tried to declare it inside of the function where it was used, however the function `GetExecutionHistoryPagesWithContext` can be called multiple times and it needs to have a reference to the same map to be able to group every set of related events into a single Job. 

@meliaj  noticed that some workflows are showing that one of the states is queued when the workflow has completed successfully. We also saw some of these logs https://github.com/Clever/workflow-manager/blob/47b1b785fdd4d52cce5c2ea930201a68f12fb555/executor/workflow_manager_sfn.go#L542

The bug prevents some Jobs within a workflow from getting updated:
Screen Shot 2019-05-06 at 2 56 35 AM

With fix:
Screen Shot 2019-05-06 at 2 47 33 AM

Note that clicking on a workflow in hubble updates the execution history and saves it, so when the fix is deployed every workflow will be updated and displayed correctly even if it ran when the buggy workflow-manager was deployed.

The `maxResults` parameter can be provided to set the limit on the number of events returned per page of results. The default is 100. 
https://docs.aws.amazon.com/step-functions/latest/apireference/API_GetExecutionHistory.html#API_GetExecutionHistory_RequestSyntax

Screen Shot 2019-05-05 at 2 48 14 PM
Note that the events corresponding to dpl-elasticsearch-sis span across a page boundary. The state entered event is 99 so it is on the first page, and the subsequent events for that Job are on the following page. Instead of creating a new eventIDToJob map each time the callback within GetExecutionHistoryPagesWithContext is called, the map should be declared in the outer scope so that it will contain a mapping for all event IDs. This prevents losing the reference to the "parent" event ID when it appears on the previous page from children events.

- [ ] Run "make generate"

### DIFF
--- a/executor/workflow_manager_sfn.go
+++ b/executor/workflow_manager_sfn.go
@@ -399,6 +399,7 @@ func (wm *SFNWorkflowManager) UpdateWorkflowHistory(ctx context.Context, workflo
 	defer cancel()
 
 	var jobs []*models.Job
+	eventIDToJob := map[int64]*models.Job{}
 	if err := wm.sfnapi.GetExecutionHistoryPagesWithContext(ctx, &sfn.GetExecutionHistoryInput{
 		ExecutionArn: aws.String(execARN),
 	}, func(historyOutput *sfn.GetExecutionHistoryOutput, lastPage bool) bool {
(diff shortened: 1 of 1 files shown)


prime-time:
I'm considering bumping up `MaxResults` since a lot of workflows have over 100 events. That should help conserve `GetExecutionHistory` calls in theory.

prime-time:
I think bumping up `maxResults` from 100 to 200 is probably ok since it's still well below the absolute limit of 1000. I'd be happy to remove that change or split it off from this pull request though.

prime-time:
I removed the commit that bumped `maxResults`

prime-time:
Wopsies, I should have done a better job proofreading the description 🙈

PR #181, opened 2019-05-08
enable tracing

prime-time:
- [ ] Run "make generate"

`LIGHTSTEP_ACCESS_TOKEN` is no longer used, so we should remove it. 

I noticed that workflow-manager logs errors on startup 
`"title": "please set TRACING_ACCESS_TOKEN to enable tracing"`
https://github.com/Clever/workflow-manager/blob/5d7b43fa0bb77050650ecea37b3ec04a7c6e8d4f/gen-go/server/router.go#L94

It seems like the other wag services have these environment variables for tracing. Let's enable tracing in workflow-manager unless there's a reason why it shouldn't be enabled.

config pull request: https://github.com/Clever/ark-config/pull/1798


prime-time:
bump

prime-time:
Is this change ok to merge? I added in a version bump, though I'm not sure if that's necessary.

mohit:
The version change is not required and my apologies for now mentioning this before but you need to change the env vars in ark-config to be able to deploy this.

prime-time:
Ah ok, I can remove the version bump. I was planning on merging both pull requests at the same time. That's what you mean, right?

mohit:
Yup

PR #184, opened 2019-05-24
add Parameters

rgarcia:
States Language now has a "Parameters" field on task states:
https://states-language.net/spec.html

Step Functions provides a Context object that can be referenced from
within Parameters using `$$`:
https://docs.aws.amazon.com/step-functions/latest/dg/input-output-contextobject.html

This PR changes our injection of _EXECUTION_NAME into the input of a
state machine with injection of via Parameters + Context. It also adds
some other values from the Context object that could be useful for
Task states.

- [x] Run "make generate"


executor/workflow_manager_sfn.go
```diff
@@ -81,30 +72,47 @@ func stateMachineWithFullActivityARNs(oldSM models.SLStateMachine, region, accou
...
-		if injectRetry && state.Type == models.SLStateTypeTask {
+		if injectRetry {
 			// Add the default retry before user defined ones since Error=States.ALL must
 			// always be the last in the retry list and could be defined in the workflow definition
-			state.Retry = append([]*models.SLRetrier{defaultSFNCLICommandTerminatedRetrier}, state.Retry...)
+			state.Retry = append([]*models.SLRetrier{sfnconventions.SFNCLICommandTerminatedRetrier}, state.Retry...)
+		}
+
+		// set up parameters
+		if state.Type == models.SLStateTypeTask {
```
Sayan-: do we still need this check here? It feels redundant since we have a `continue` earlier in the loop: https://github.com/Clever/workflow-manager/pull/184/files?file-filters%5B%5D=.go#diff-f6849ad311da09293f15ce0693474f50R84

executor/sfnconventions/sfnconventions.go
```diff
@@ -68,3 +71,26 @@ func EmbeddedResourceArn(wdResource, region, accountID, namespace string, app st
...
+// these errors.
+var SFNCLICommandTerminatedRetrier = &models.SLRetrier{
+	BackoffRate:     1.0,
+	ErrorEquals:     []models.SLErrorEquals{SFNCLICommandTerminated},
+	IntervalSeconds: 10,
+	MaxAttempts:     swag.Int64(10),
+}
+
+// TaskStateParameters pulls in Context values set by from AWS Step Functions into a Task state's Parameters object.
+// https://docs.aws.amazon.com/step-functions/latest/dg/input-output-contextobject.html
```
Sayan-: this is neat!

executor/sfnconventions/sfnconventions.go
```diff
@@ -68,3 +71,26 @@ func EmbeddedResourceArn(wdResource, region, accountID, namespace string, app st
...
+var TaskStateParameters = map[string]interface{}{
+	"_EXECUTION_NAME":     "$$.Execution.Name",
```
natebrennand: should we note that this is a UUID in WFM world?

executor/workflow_manager_sfn.go
```diff
@@ -81,30 +72,47 @@ func stateMachineWithFullActivityARNs(oldSM models.SLStateMachine, region, accou
...
+			if state.Parameters == nil {
+				state.Parameters = map[string]interface{}{}
+			}
+			for k, v := range sfnconventions.TaskStateParameters {
```
mohit: Based on a quick reading it seems like for dynamic properties pulled from the context there is a .$ at the end of the key: https://docs.aws.amazon.com/step-functions/latest/dg/input-output-inputpath-params.html

mohit:
Maybe we should text this change in dev with a data-pipeline and multiverse workflow since an issue here could cause a big backlog :)

executor/workflow_manager_sfn.go
rgarcia: derp yes 99153a1

executor/sfnconventions/sfnconventions.go
rgarcia: :nods: 514be5f

executor/workflow_manager_sfn.go
rgarcia: yes, nice catch! 1c20af6

rgarcia:
Yes, planning to test in dev

rgarcia:
bad news: it looks like `Parameters` completely overrides the input to a state 😢 

going to close this for now

PR #208, opened 2019-08-08
optionally omit execution history + use lastJob input when rerunning workflows

prime-time:
This pull request contains two changes:
- Introduce `omitExecutionHistory` as an optional query parameter for the `GetWorkflowByID` API endpoint.
- Allow the `ResumeWorkflowByID` endpoint to use the input payload from `lastJob` when available. 

Returning a workflow object with its `jobs` array populated requires querying the `GetExecutionHistory` endpoint via the stepfunctions API which is a relatively expensive operation. Sometimes workflows get stuck in a loop and accumulate so many execution events that the `GetExecutionHistory` calls can time out causing the request to fail. The `omitExecutionHistory` allows us to fetch the workflow summary when the execution history is long without failing, and it allows us to fetch workflows more efficiently when the jobs array isn't needed. 
Calling the `GetWorkflowByID` endpoint with `omitExecutionHistory` set to `true` will return the workflow object without a jobs array.

`ResumeWorkflowByID` will now use the input from `lastJob` when available instead of only searching through the jobs array. `lastJob` is populated for failed workflows, and it populated automatically when the workflow update loop calls `UpdateWorkflowSummary`. Prior to this change, the only way to rerun a workflow was by first making a separate `GetWorkflowByID` call to populate the jobs array. We almost always want to rerun workflows from the state in which they failed, and this change allows workflows to be rerun without needing the full execution history. It's still possible to rerun a workflow from an earlier state in the state machine, but that requires pre-populating the jobs array as before.

Now it's possible to rerun a workflow even if its execution history cannot be loaded.

I tested this is staging by querying `GetWorkflowByID` with `omitExecutionHistory`, and I also tested `ResumeWorkflowByID` through the js client. 

I reran 1000 workflows with and without these changes, and I've included screenshots of the dynamodb metrics below.

Rerunning workflows without these changes results in dynamodb throttling.
Screen Shot 2019-08-08 at 8 13 48 AM

With these changes, fewer dynamodb writes are consumed. 
Screen Shot 2019-08-08 at 1 36 53 PM

- [x] Run "make generate"

### DIFF
--- a/docs/overview.md
+++ b/docs/overview.md
@@ -7,7 +7,7 @@ Orchestrator for AWS Step Functions
 
 
 ### Version information
-*Version* : 0.10.0
+*Version* : 0.11.0
 
 
 ### URI scheme
--- a/docs/paths.md
+++ b/docs/paths.md
@@ -295,9 +295,10 @@ GET /workflows/{workflowID}
 
 #### Parameters
 
-|Type|Name|Schema|
-|---|---|---|
-|**Path**|**workflowID**  <br>*required*|string|
+|Type|Name|Description|Schema|Default|
+|---|---|---|---|---|
+|**Path**|**workflowID**  <br>*required*||string||
+|**Query**|**omitExecutionHistory**  <br>*optional*|Skips fetching the full execution history, and omits the jobs array.|boolean|`"false"`|
 
 
 #### Responses
--- a/embedded/embedded.go
+++ b/embedded/embedded.go
@@ -430,7 +430,8 @@ func (e *Embedded) CancelWorkflow(ctx context.Context, i *models.CancelWorkflowI
 }
 
 // GetWorkflowByID ...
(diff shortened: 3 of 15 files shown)


kvconfig.yml
```diff
@@ -1,6 +1,6 @@
 ---
 routes:
-  # workflow-manager related matchers; prefix with workflow-manager
+  # workflow-manager related matchers; prefix with workflow-mananger
```
ghirsch1: This was right before.

prime-time: what was I thinking...

alsmola:
What's the status on this PR? Should we merge?

prime-time:
Ah I forgot to update the status after I opened this. I spoke with @mohit about it in person and on slack, and he recommended that we explore a different approach that doesn't involve introducing a query parameter. I started working on a separate branch, but haven't had a chance to finish it yet.

PR #211, opened 2019-08-30
add metric for failed workflows with dimension for failing job

ghirsch1:
https://clever.atlassian.net/browse/INFRA-3435


executor/workflow_manager_sfn.go
```diff
@@ -392,6 +392,16 @@ func (wm *SFNWorkflowManager) UpdateWorkflowSummary(ctx context.Context, workflo
 		if err := wm.updateWorkflowLastJob(ctx, workflow); err != nil {
 			return err
 		}
+		failedJob := ""
+		if workflow.LastJob != nil {
+			failedJob = workflow.LastJob.State
```
mohit: you probably want both the stateName and the Resource name since State is often set to `start`. Adding a Resource name would allow teams to alert on failures for a specific worker (aka resource) across workers.

PR #212, opened 2019-09-03
Use the input from lastJob when it's available

prime-time:
This is a second attempt at https://github.com/Clever/workflow-manager/pull/208

Given a workflow ID and the name of a workflow state to start from, `resumeWorkflowByID` creates a new workflow that begins execution from the `StartAt` state and uses the input to that state from the original workflow. The new workflow is then linked to the original as a retry. The way that `resumeWorkflowByID` finds the workflow input that it needs is by searching through the original workflow's `jobs` array. It relies on `jobs` being populated ahead of time, however `jobs` is not populated automatically like other workflow fields are. In order for `jobs` to exist for a workflow, `getWorkflowByID` must be called before `resumeWorkflowByID` is called, otherwise `jobs` will be `nil` and `resumeWorkflowByID` will error. 

This pull request helps solve a couple of issues: 
- A workflow with an extremely long execution history cannot be rerun because its `jobs` array can't be populated.
- Rerunning many workflows at the same time can result in dynamo db throttling because of the needed `getWorkflowByID` calls.

With this change, `resumeWorkflowByID` uses the workflow input from `lastJob` when possible. If `lastJob` is not available, then it will load the full `jobs` array and search through it to find the input that it needs. `resumeWorkflowByID` is almost always used to rerun a workflow from the state in which it failed, so in practice the input will be taken from `lastJob` a vast majority of the time. By using `lastJob` instead of `jobs`, we avoid needing to populate `jobs` and we avoid lots of dynamodb writes.

I tested this by running workflow-manager locally and rerunning failed workflows.

- [ ] Run "make generate"

### DIFF
--- a/handler.go
+++ b/handler.go
@@ -360,31 +360,55 @@ func (h Handler) ResumeWorkflowByID(ctx context.Context, input *models.ResumeWor
 	if !resources.WorkflowIsDone(&workflow) {
 		return &models.Workflow{}, fmt.Errorf("Workflow %s active: %s", workflow.ID, workflow.Status)
 	}
+	isDone, err := descendantsAreDone(ctx, h.store, workflow)
+	if err != nil {
+		return &models.Workflow{}, fmt.Errorf("Failed to check descendant workflows: %s", err)
+	}
+	if !isDone {
+		return &models.Workflow{}, fmt.Errorf("Workflow %s has active descendants: %s", workflow.ID, workflow.Status)
+	}
 	if _, ok := workflow.WorkflowDefinition.StateMachine.States[input.Overrides.StartAt]; !ok {
 		return &models.Workflow{}, fmt.Errorf("Invalid StartAt state %s", input.Overrides.StartAt)
 	}
 
-	// find the input to the StartAt state
-	effectiveInput := ""
-	for _, job := range workflow.Jobs {
-		if job.State == input.Overrides.StartAt {
-			// if job was never started then we should probably not trust the input
-			if job.Status == models.JobStatusAbortedDepsFailed ||
-				job.Status == models.JobStatusQueued ||
-				job.Status == models.JobStatusWaitingForDeps ||
-				job.Status == models.JobStatusCreated {
-
-				return &models.Workflow{},
-					fmt.Errorf("Job %s for StartAt %s was not started for Workflow: %s. Could not infer input",
-						job.ID, job.State, workflow.ID)
+	// Find the job that ran for the StartAt state so that its input can be used for the new workflow.
+	// Try to use lastJob if it's available and if we want to start the new workflow from that state,
+	// otherwise load the full execution history and search through it to find the desired input.
+	var originalJob *models.Job
+	if workflow.LastJob != nil && input.Overrides.StartAt == workflow.LastJob.State {
+		originalJob = workflow.LastJob
+	} else {
+		if workflow.Jobs == nil {
+			if err := h.manager.UpdateWorkflowHistory(ctx, &workflow); err != nil {
+				return &models.Workflow{}, err
 			}
+			updatedWorkflow, err := h.store.GetWorkflowByID(ctx, input.WorkflowID)
+			if err != nil {
+				return &models.Workflow{}, err
+			}
+			workflow = updatedWorkflow
+		}
 
-			effectiveInput = job.Input
-			break
+		for _, job := range workflow.Jobs {
+			if job.State == input.Overrides.StartAt {
+				originalJob = job
+				break
+			}
 		}
 	}
 
-	return h.manager.RetryWorkflow(ctx, workflow, input.Overrides.StartAt, effectiveInput)
+	if originalJob == nil {
+		return &models.Workflow{}, fmt.Errorf("No job found for StartAt %s", input.Overrides.StartAt)
+	}
+
+	// if job was never started then we should probably not trust the input
+	if !hasJobStarted(originalJob.Status) {
+		return &models.Workflow{},
+			fmt.Errorf("Job %s for StartAt %s was not started for Workflow: %s. Could not infer input",
+				originalJob.ID, originalJob.State, workflow.ID)
+	}
+
+	return h.manager.RetryWorkflow(ctx, workflow, input.Overrides.StartAt, originalJob.Input)
 }
 
 // ResolveWorkflowByID sets a workflow's ResolvedByUser to true if it is currently false.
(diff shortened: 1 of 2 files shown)


handler.go
```diff
@@ -360,31 +360,55 @@ func (h Handler) ResumeWorkflowByID(ctx context.Context, input *models.ResumeWor
...
-				job.Status == models.JobStatusQueued ||
-				job.Status == models.JobStatusWaitingForDeps ||
-				job.Status == models.JobStatusCreated {
-
-				return &models.Workflow{},
-					fmt.Errorf("Job %s for StartAt %s was not started for Workflow: %s. Could not infer input",
-						job.ID, job.State, workflow.ID)
+	// Find the job that ran for the StartAt state so that its input can be used for the new workflow.
+	// Try to use lastJob if it's available and if we want to start the new workflow from that state,
+	// otherwise load the full execution history and search through it to find the desired input.
```
Sayan-: I would note the `jobs == nil` case is expensive. It might be worth noting clients of this implementation might want to increase the timeout from the global default

handler.go
```diff
@@ -360,31 +360,55 @@ func (h Handler) ResumeWorkflowByID(ctx context.Context, input *models.ResumeWor
...
+	var originalJob *models.Job
+	if workflow.LastJob != nil && input.Overrides.StartAt == workflow.LastJob.State {
+		originalJob = workflow.LastJob
+	} else {
```
Sayan-: TODO - add `input.Overrides.StartAt == workflow.StartAt` short circuit

handler.go
```diff
@@ -443,3 +474,24 @@ func validateTagsMap(apiTags map[string]interface{}) error {
 func epochMillis(t time.Time) int {
 	return int(t.UnixNano() / int64(time.Millisecond))
 }
+
+// descendantsAreDone checks a workflow's descendants for any active workflows.
```
Sayan-: nit - I'd add a note this is doing a depth first search so a reader doesn't need to look at the implementation

handler_test.go
```diff
@@ -83,3 +113,351 @@ func TestStartWorkflow(t *testing.T) {
...
+
+	specs := []struct {
+		desc    string
+		isError bool
+		startAt string
+		wf      *models.Workflow
+		newWF   *models.Workflow
+	}{
+		{
+			desc: "uses the input from lastJob if StartAt == lastJob.State" +
```
Sayan-: any reason to not make this oneline?

handler_test.go
```diff
@@ -83,3 +113,351 @@ func TestStartWorkflow(t *testing.T) {
 		assert.NoError(t, err)
 	}
 }
+
+func TestResumeWorkflowByID(t *testing.T) {
```
Sayan-: TODO - test for start short circuit

handler.go
```diff
@@ -360,31 +360,55 @@ func (h Handler) ResumeWorkflowByID(ctx context.Context, input *models.ResumeWor
 	if !resources.WorkflowIsDone(&workflow) {
 		return &models.Workflow{}, fmt.Errorf("Workflow %s active: %s", workflow.ID, workflow.Status)
 	}
+	isDone, err := descendantsAreDone(ctx, h.store, workflow)
+	if err != nil {
+		return &models.Workflow{}, fmt.Errorf("Failed to check descendant workflows: %s", err)
+	}
+	if !isDone {
+		return &models.Workflow{}, fmt.Errorf("Workflow %s has active descendants: %s", workflow.ID, workflow.Status)
```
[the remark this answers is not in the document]
prime-time: I was actually going with the vernacular that hubble uses for this check. `descendants` is a bit more accurate since it's checking children at all levels and not just direct children. 
`descendantWorkflows` sounds pretty good. We could also go with `descendantRetries` or just `retries`.

handler.go
prime-time: 💸

handler.go
prime-time: Good idea

handler.go
```diff
@@ -430,6 +454,13 @@ func newWorkflowDefinitionFromRequest(req models.NewWorkflowDefinitionRequest) (
 	return resources.NewWorkflowDefinition(req.Name, req.Manager, req.StateMachine, req.DefaultTags)
 }
 
+func hasJobStarted(status models.JobStatus) bool {
```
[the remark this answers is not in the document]
prime-time: Hm good question. I wonder if we can omit this check entirely. There haven't been any errors in the past 30 days, and I wonder if it's ever occurred.

handler.go
prime-time: In this case, none of the execution history would be loaded so we wouldn't be able to do any of the checks on the job's status.

https://github.com/Clever/workflow-manager/blob/3bd2e478da6287a2983d575f965ff010905b86ce/handler.go#L372-L375

I wonder if we could just delete these checks. I'll have to think about it some more.

handler_test.go
Sayan-: Ahhh - I don't think we have a linting rule for that in go; bnd

handler.go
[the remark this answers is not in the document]
Sayan-: sounds good - thanks for clarifying!

handler.go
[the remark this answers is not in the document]
Sayan-: 🤔 I'm honestly not familiar enough with this path to provide helpful feedback. Might be worth a quick chat with mo

handler.go
prime-time: To get the first job, we could call `GetExecutionHistory` once and not page through to the end.

prime-time: Actually let's skip this optimization for now, and we can revisit it if we need to. I think it's best to keep it simple, and it that optimization doesn't apply to most of our use cases for this api endpoint.

This new implementation only calls `GetExecutionHistory` when `jobs` is empty, and for that case the `GetWorkflowByID` endpoint can be called beforehand to populate `jobs`. That's how `ResumeWorkflowByID` worked before, so it would be no more inefficient.

handler.go
prime-time: b15d65b

handler.go
prime-time: b28d09d

handler.go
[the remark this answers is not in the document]
prime-time: Ok I can check with mo

handler.go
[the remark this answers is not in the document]
prime-time: I did some minor renaming here b15d65b

handler_test.go
prime-time: ⚡️ 🔌 🐻

handler_test.go
```diff
@@ -83,3 +113,351 @@ func TestStartWorkflow(t *testing.T) {
...
+	workflowDefinition := &models.WorkflowDefinition{
+		StateMachine: &models.SLStateMachine{
+			StartAt: "monkey-state",
+			States: map[string]models.SLState{
+				"monkey-state": models.SLState{
+					Type:     models.SLStateTypeTask,
+					Next:     "gorilla-state",
+					Resource: "resource-name",
+				},
+				"gorilla-state": models.SLState{
```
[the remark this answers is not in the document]
prime-time: 🍌

prime-time:
@Sayan- How does this look with the latest changes? I think I addressed all the feedback.

PR #213, opened 2019-09-03
Workflow Manager Tracing

taylor-sutton:
This branch adds tracing to workflow manager (in addition to the baseline tracing from wag). It uses the same `opentracing-contrib/aws-sdk-go` to put some coarse tracing around SFN and SQS calls, uses some handwritten spans for elasticsearch, and throws in a couple of spans around some other things for good measure. In order to make this work well, I needed to pass around a context in a few more places than already had it. 

### Remarks
* In one place, I changed an existing log call to use a different key name - from "wf-id" to "workflow-id" which is used in other places within this library and elsewhere. I don't think this breaks anything, but it would make searching for logs across a bit weird. It can still be done with `wf-id:<ID> OR workflow-id:<ID>`, though, so I think I'd rather have it cleaner going forward.

* I'm _fairly_ sure passing a context to the AWS calls doesn't affect functionality, but I'm not 100% sure. 

* Probably this shouldn't be the final home for the little elasticsearch tracing wrapper, but maybe it's okay to leave it here for now as a proof of concept? Unless you have a suggestion for somewhere better!

### PR Template Checklist:

- [x] Update swagger.yml version - no changes, so not needed!
- [x] Run "make generate"

### DIFF
--- a/executor/workflow_manager_sfn.go
+++ b/executor/workflow_manager_sfn.go
@@ -125,10 +125,11 @@ func toSFNTags(wmTags map[string]interface{}) []*sfn.Tag {
 	return sfnTags
 }
 
-func (wm *SFNWorkflowManager) describeOrCreateStateMachine(wd models.WorkflowDefinition, namespace, queue string) (*sfn.DescribeStateMachineOutput, error) {
-	describeOutput, err := wm.sfnapi.DescribeStateMachine(&sfn.DescribeStateMachineInput{
-		StateMachineArn: aws.String(sfnconventions.StateMachineArn(wm.region, wm.accountID, wd.Name, wd.Version, namespace, wd.StateMachine.StartAt)),
-	})
+func (wm *SFNWorkflowManager) describeOrCreateStateMachine(ctx context.Context, wd models.WorkflowDefinition, namespace, queue string) (*sfn.DescribeStateMachineOutput, error) {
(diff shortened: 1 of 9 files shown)


main.go
```diff
@@ -185,3 +197,46 @@ func logSFNCounts(sfnCounter *counter.Counter) {
...
+func (t tracedESTransport) Perform(req *http.Request) (*http.Response, error) {
+	span, ctx := opentracing.StartSpanFromContext(req.Context(), "elasticsearch-request")
+	defer span.Finish()
+
+	// These fields mirror the ones in the aws-sdk-go opentracing package
+	ext.SpanKindRPCClient.Set(span)
+	ext.Component.Set(span, "go-elasticsearch")
+	ext.HTTPMethod.Set(span, req.Method)
+	ext.HTTPUrl.Set(span, req.URL.String())
+	ext.PeerService.Set(span, "elasticsearch")
```
natebrennand: 👏

executor/workflow_manager_sfn.go
```diff
@@ -415,8 +417,7 @@ func (wm *SFNWorkflowManager) UpdateWorkflowHistory(ctx context.Context, workflo
 
 	// Setup a context with a timeout of one minute since
 	// we don't want to pull very large workflow histories
-	// TODO: this should be a context passed by the handler
```
natebrennand: 👌

PR #214, opened 2019-09-03
Tracing cleanup

taylor-sutton:
After looking at the source for `http.TimeoutHandler`, I think it does everything out custom little middleware gizmo does, so we don't need our custom one. Also, add some tagging to `updateWorkflow` with the result of the operation, so it can be filtered more easily.

- [x] Run "make generate"

### DIFF
--- a/executor/workflow_manager.go
+++ b/executor/workflow_manager.go
@@ -123,15 +123,18 @@ func updatePendingWorkflow(ctx context.Context, m *sqs.Message, wm WorkflowManag
 			// workflow has disappeared from our DB. No sense in
 			// trying to update it again, so delete the SQS message
 			deleteMsg()
-			return "", fmt.Errorf("worfklow id not found: %s", wfID)
+			span.SetTag("result", "workflow-not-found")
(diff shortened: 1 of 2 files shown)


executor/workflow_manager.go
```diff
@@ -123,11 +123,13 @@ func updatePendingWorkflow(ctx context.Context, m *sqs.Message, wm WorkflowManag
...
 			// trying to update it again, so delete the SQS message
 			deleteMsg()
-			return "", fmt.Errorf("worfklow id not found: %s", wfID)
+			span.SetTag("result", "workflow-not-found")
+			return "", fmt.Errorf("workflow id not found: %s", wfID)
 		}
 		// other error, e.g. throttling. Try again later
 		deleteMsg()
 		requeueMsg()
+		span.SetTag("result", "database-error")
```
natebrennand: do you wanna save the database error message too?

taylor-sutton: The error gets returned by this function and then logged by the function executing the polling loop. So maybe it's okay like this? I'll admit I'm not really sure how much duplication there should be between tracing and logs.

PR #215, opened 2019-09-04
Infra 3428 startworkflow failures

taylor-sutton:
Jira reference: https://clever.atlassian.net/browse/INFRA-3482

This way, every workflow that was started will be picked up by an update loop, getting around the problem of a workflow not being updated because something went wrong before the SQS message was sent. See the discussion on Jira for more info.

- [x] Update swagger.yml version - not needed
- [x] Run "make generate"

### DIFF
--- a/executor/workflow_manager.go
+++ b/executor/workflow_manager.go
@@ -122,6 +122,9 @@ func updatePendingWorkflow(ctx context.Context, m *sqs.Message, wm WorkflowManag
 		if _, ok := err.(models.NotFound); ok {
 			// workflow has disappeared from our DB. No sense in
 			// trying to update it again, so delete the SQS message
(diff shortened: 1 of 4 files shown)


taylor-sutton:
Totally right about RetryWorkflow, I will update it! Good catch.

Sayan-:
One last thing I'd like to push on here - we're following a `delete` -> `requeue` pattern in: https://github.com/Clever/workflow-manager/blob/57e47fcb2fee3e2c0875a7c4e72f09de8298f808/executor/workflow_manager.go#L97-L154

I wonder if the better option here is to `requeue` -> `delete` to ensure we have _at least one_ copy of the message. Love the simplicity of this fix!

PR #220, opened 2019-10-01
Add context to cached sfnapi

aastein:
- [ ] Run "make generate"

### DIFF
--- a/executor/sfncache/sfn.go
+++ b/executor/sfncache/sfn.go
@@ -27,7 +27,7 @@ func New(sfnapi sfniface.SFNAPI) (sfniface.SFNAPI, error) {
 	}, nil
 }
 
-// examplee: arn:aws:states:us-west-1:589690932525:stateMachine:production--clever-dev--launchpad.d259---1--Start
+// examplee: arn:aws:states:us-west-1:123:stateMachine:env--org--appdeploy---1--Start
 var arnVersionRegex = regexp.MustCompile(`.*:stateMachine:.+--.+---1--.+`)
 
 // DescribeStateMachineWithContext is cached aggressively since state machines are immutable.
--- a/executor/sfncache/sfn.go
+++ b/executor/sfncache/sfn.go
@@ -27,7 +27,7 @@ func New(sfnapi sfniface.SFNAPI) (sfniface.SFNAPI, error) {
 	}, nil
 }
 
-// examplee: arn:aws:states:us-west-1:123:stateMachine:env--org--appdeploy---1--Start
+// example: arn:aws:states:us-west-1:123:stateMachine:env--org--appdeploy---1--Start
 var arnVersionRegex = regexp.MustCompile(`.*:stateMachine:.+--.+---1--.+`)
 
 // DescribeStateMachineWithContext is cached aggressively since state machines are immutable.
--- a/executor/sfncache/sfn.go
+++ b/executor/sfncache/sfn.go
@@ -30,7 +30,7 @@ func New(sfnapi sfniface.SFNAPI) (*SFNCache, error) {
 // example: arn:aws:states:us-west-1:123:stateMachine:env--org--appdeploy---1--Start
 var arnVersionRegex = regexp.MustCompile(`.*:stateMachine:.+--.+---1--.+`)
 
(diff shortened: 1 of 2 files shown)


executor/sfncache/sfn.go
```diff
@@ -1,39 +1,42 @@
...
-		SFNAPI:                    sfnapi,
-		describeStateMachineCache: describeStateMachineCache,
+		SFNAPI:                               sfnapi,
+		describeStateMachineWithContextCache: describeStateMachineCacheWithContext,
 	}, nil
 }
 
-// DescribeStateMachine is cached aggressively since state machines are immutable.
-func (s *SFNCache) DescribeStateMachine(i *sfn.DescribeStateMachineInput) (*sfn.DescribeStateMachineOutput, error) {
+// DescribeStateMachineWithContext is cached aggressively since state machines are immutable.
```
Sayan-: Actually - we allow for mutable state machines now....we should add a `version != -1` check before adding a key to the cache

aastein: idk SFN stuff. not seeing a version field in the output to check against?

Sayan-: We _could_ add a method to parse the version out of the state machine name. Essentially reverse this bit:

https://github.com/Clever/workflow-manager/blob/1d8e7da9d096a48f04045c8f085e8c72941f54a3/executor/workflow_manager_sfn.go#L131

https://github.com/Clever/workflow-manager/blob/1d8e7da9d096a48f04045c8f085e8c72941f54a3/executor/sfnconventions/sfnconventions.go#L47-L50

so we don't break the sfnapi...

I'm not actually sure if we use this feature outside of the `embedded workflow-manager` use case.....

aastein: sketchy. lmk if what I just pushed is what you mean.

Sayan-: regex or a string parse is fine!

can we just add a test or two that use the underlying `sfnconventions.StateMachineArn` call 🙈

executor/sfncache/sfn.go
```diff
@@ -1,39 +1,50 @@
...
-		SFNAPI:                    sfnapi,
-		describeStateMachineCache: describeStateMachineCache,
+		SFNAPI:                               sfnapi,
+		describeStateMachineWithContextCache: describeStateMachineCacheWithContext,
 	}, nil
 }
 
-// DescribeStateMachine is cached aggressively since state machines are immutable.
-func (s *SFNCache) DescribeStateMachine(i *sfn.DescribeStateMachineInput) (*sfn.DescribeStateMachineOutput, error) {
+// examplee: arn:aws:states:us-west-1:589690932525:stateMachine:production--clever-dev--launchpad.d259---1--Start
```
Sayan-: this is technically a public repo, let's snip out any clever specific stuff

aastein: woops, TIL

executor/sfncache/sfn.go
```diff
@@ -1,39 +1,50 @@
...
-		describeStateMachineCache: describeStateMachineCache,
+		SFNAPI:                               sfnapi,
+		describeStateMachineWithContextCache: describeStateMachineCacheWithContext,
 	}, nil
 }
 
-// DescribeStateMachine is cached aggressively since state machines are immutable.
-func (s *SFNCache) DescribeStateMachine(i *sfn.DescribeStateMachineInput) (*sfn.DescribeStateMachineOutput, error) {
+// example: arn:aws:states:us-west-1:123:stateMachine:env--org--appdeploy---1--Start
+var arnVersionRegex = regexp.MustCompile(`.*:stateMachine:.+--.+---1--.+`)
```
rgarcia: took me a minute to catch this, but I'd comment that this involves two conventions of ours:
1. our naming convention for state machines (perhaps this code could rely on `sfnconventions` package instead of regex?)
2. setting the version component of that name to `-1` for mutable state machines

rgarcia: also I'm not super convinced we need to be this precise and miss the cache in this case, since the only place we use this convention is in embedded workflow manager, which doesn't use sfncache anyway

PR #222, opened 2019-10-09
adding pod shortname to tracing tags

ghirsch1:


ghirsch1:
duped in the meantime

PR #223, opened 2019-10-10
Embedded allow setting execution name

Sayan-:
This PR attempts to address the problem that embedded `workflow-manager` execution names do not encode human readable information. The approach we use here is to allow clients to pass in a `suffix` we append and trim to be within step functions limits

I don't think we're making any breaking changes by switching away from a UUID. That field isn't actually used in the embedded pathway at all.

My main concern is in regard to Step Functions consistency: https://docs.aws.amazon.com/step-functions/latest/dg/concepts-read-consistency.html

right now we have a 5s delay between `Update` and `StartExecution`

### DIFF
--- a/docs/definitions.md
+++ b/docs/definitions.md
@@ -219,12 +219,13 @@
 <a name="startworkflowrequest"></a>
 ### StartWorkflowRequest
 
-|Name|Schema|
-|---|---|
-|**input**  <br>*optional*|string|
-|**namespace**  <br>*optional*|string|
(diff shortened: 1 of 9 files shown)


rgarcia:
I think it's good but I'd go even further and allow users to specify the full ID. It's common to see client generated IDs (e.g. databases allow this) and would clean up places where we need to store the workflow id (dapple has some gnarly logic to capture the generated workflow id). For human readable text we could support tagging workflows

PR #224, opened 2019-10-10
Add some uniqueness to suffixed workflow ids

Sayan-:


taylor-sutton:
If my math is right, with 5 random hex digits, there's a 1% chance that there will be a conflict in a given app within 150 deploys [1]. With 100 apps, each with 150 deploys, there's a 64% chance that at least one app conflicts. [2]

Those numbers of deploys and of apps might be a little high, but I think 5 digits is probably too few if I'm understanding right that they are just hex digits.

Have we made explicit anywhere the constraints that arise from embedded workflow names?

1: https://www.wolframalpha.com/input/?i=product+of+%2816%5E5-n%29%2F16%5E5+from+n%3D0+to+150
2: 1-.99^100

PR #255, opened 2020-08-28
glue jobs in workflows

ghirsch1:
## Link to JIRA:
https://clever.atlassian.net/browse/INFRANG-4008

## Overview:
Adding support for Glue jobs in workflows. They'll look like this
```
States:
  start:
    End: true
    Parameters:
      Arguments.$: $.PATH_NAME
    Resource: glue:APP_NAME
    TimeoutSeconds: 7200
    Type: Task
```

This gets transformed into something like
```
States:
  start:
    End: true
    Parameters:
      JobName: ENVIRONMENT/NAMESPACE--APP_NAME
      Arguments.$: $.PATH_NAME
    Resource: arn:aws:states:::glue:startJobRun.sync
    TimeoutSeconds: 7200
    Type: Task
```

To pass in arguments, you can either specify them explicitly, or to pull them from the path you can do it like above with the `.$/$.` notations

- [x] Run "make generate"

## Testing
- [Published a new workflow definition in dev](https://us-west-1.console.aws.amazon.com/dynamodb/home?region=us-west-1#tables:selected=workflow-manager-dev-v3-latest-workflow-definitions;tab=items)
- [Successful execution submitted by wfm in dev](https://us-west-2.console.aws.amazon.com/states/home?region=us-west-2#/executions/details/arn:aws:states:us-west-2:589690932525:execution:clever-dev--test-glue-workflow--1--start:ee63af16-297d-4070-8f44-a9b25d59e0ca)

## Rollout
- [ ] announce to #eng and #eng-analytics and #data-engineering
- [x] add some details to https://clever.atlassian.net/wiki/spaces/ENG/pages/246579345/Workflows and https://clever.atlassian.net/wiki/spaces/ENG/pages/754745733/Glue
- [ ] add default workflow definition for glue template in init-service

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,2 +1,13 @@
+## Link to JIRA:
+[Link to JIRA](https://clever.atlassian.net/browse/INFRANG-????)
+
+## Overview:
+[insert description]
+
+If you made any changes to swagger.yml:
 - [ ] Update swagger.yml version
 - [ ] Run "make generate"
+
+## Testing
+
+## Rollout
--- a/executor/workflow_manager_sfn.go
+++ b/executor/workflow_manager_sfn.go
@@ -26,6 +26,8 @@ const (
 	maxFailureReasonLines   = 3
 	executionEventsPerPage  = 200
 	sfncliCommandTerminated = "sfncli.CommandTerminated"
+
+	jobNameField = "JobName"
 )
 
 var durationToRetryDescribeExecutions = 5 * time.Minute
@@ -61,10 +63,10 @@ func NewSFNWorkflowManager(sfnapi sfniface.SFNAPI, sqsapi sqsiface.SQSAPI, store
 	}
 }
 
-// stateMachineWithFullActivityARNs converts resource names in states to full activity ARNs. It returns a new state machine.
-// Our workflow definitions contain state machine definitions with short-hand for resource names, e.g. "Resource": "name-of-worker"
-// Convert this shorthand into a new state machine with full activity ARNs, e.g. "Resource": "arn:aws:states:us-west-2:589690932525:activity:production--name-of-worker"
-func stateMachineWithFullActivityARNs(oldSM models.SLStateMachine, region, accountID, namespace string) *models.SLStateMachine {
+// stateMachineWithFullActivityARNsAndParameters returns a new state machine.
+// Our workflow definitions contain state machine definitions with short-hand for resource names and parameter fields, e.g. "Resource": "name-of-worker"
+// Convert this shorthand into a new state machine with full details, e.g. "Resource": "arn:aws:states:us-west-2:589690932525:activity:production--name-of-worker"
+func stateMachineWithFullActivityARNsAndParameters(oldSM models.SLStateMachine, region, accountID, namespace string) *models.SLStateMachine {
 	sm := deepcopy.Copy(oldSM).(models.SLStateMachine)
 	for stateName, s := range sm.States {
 		state := deepcopy.Copy(s).(models.SLState)
@@ -73,6 +75,13 @@ func stateMachineWithFullActivityARNs(oldSM models.SLStateMachine, region, accou
 		}
 		if strings.HasPrefix(state.Resource, "lambda:") {
 			state.Resource = sfnconventions.LambdaResource(state.Resource, region, accountID, namespace)
+		} else if strings.HasPrefix(state.Resource, "glue:") {
+			var jobName string
+			state.Resource, jobName = sfnconventions.GlueResourceAndJobName(state.Resource, namespace)
+			if state.Parameters == nil {
+				state.Parameters = map[string]interface{}{}
+			}
+			state.Parameters[jobNameField] = jobName
 		} else {
 			state.Resource = sfnconventions.SFNCLIResource(state.Resource, region, accountID, namespace)
 		}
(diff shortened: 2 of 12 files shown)


rgarcia:
> The ArgumentsPath field is required and must look like $.PATH_NAME. And PATH_NAME must be included in the input to the job (i.e. outputted by one of the previous states).

I'm confused... can you go into more detail, maybe with some examples? E.g. I can think of two cases
- I want to trigger the job with the default arguments
- I want to trigger the job with some arguments overridden

ghirsch1:
Sorry for being unclear. As an example with the test workflow that I ran, I set `ArgumentsPath: $.GlueParameters` which when converted to "real" notation looks like
```
Parameters:
  Arguments.$: $.GlueParameters
```

This means that when submitting the job, Step Functions will populate the arguments from `$.GlueParameters` on the IOPath. Since the test workflow that I was using had only the one state, I passed in as input to the function
```
{
  "GlueArguments": {
    "--day": "01",
    "--district": "57bc7923f7f2480100000930",
    "--location": "US/Central",
    "--month": "08",
    "--user-type": "schooladmins",
    "--year": "2020"
  }
}
```

which added those arguments to the default arguments (and if any had already been specified as default arguments, would have overriden them)

Now that you mention it though, maybe we would sometimes want to trigger workflows where we don't need to add/override any arguments. So it's probably better to not REQUIRE that `ArgumentsPath` field in our shorthand, and instead just make sure that it's formatted correctly if it is included.

Hopefully that clears things up, but I'm happy to discuss further.

rgarcia:
Yes, that clears things up! Is there a reason to make it `ArgumentsPath` instead of `Arguments.$`? Generally speaking the schema for workflow definitions has matched the states language schema 1-1, and in processing it we do transformations of values only, and sometimes add fields. E.g. transforming values like `glue:APP_NAME` -> `arn:...` or adding default retry policies. As written I think `ArgumentsPath` would be the first foray into augmenting the states language schema.

.github/pull_request_template.md
```diff
@@ -1,2 +1,13 @@
+## Link to JIRA:
+[Link to JIRA](url)
```
rgarcia, line 2:
```suggestion
[Link to JIRA](https://clever.atlassian.net/browse/INFRANG-????)
```

ghirsch1:
Yeah you're right, it also seems much cleaner to not do any replacements there. Plus it allows people to specify arguments directly if they want to do that instead of putting them in the path.

I think it was just one of those things where I iterated through a few ways of structuring things but forgot to reevaluate again at the end.

I'll update the PR and let you know when it's ready for review.

executor/workflow_manager_sfn.go
```diff
@@ -73,12 +75,25 @@ func stateMachineWithFullActivityARNs(oldSM models.SLStateMachine, region, accou
 		}
 		if strings.HasPrefix(state.Resource, "lambda:") {
 			state.Resource = sfnconventions.LambdaResource(state.Resource, region, accountID, namespace)
+		} else if strings.HasPrefix(state.Resource, "glue:") {
+			if state.Parameters == nil || state.Parameters[argumentsPathField] == nil {
+				return nil, fmt.Errorf("glue task state must have parameter ArgumentsPath")
```
rgarcia, line 80:
```suggestion
				return nil, fmt.Errorf("glue task state must have parameter %s", argumentsPathField)
```

executor/workflow_manager_sfn.go
```diff
@@ -73,12 +75,25 @@ func stateMachineWithFullActivityARNs(oldSM models.SLStateMachine, region, accou
...
+			}
+			argumentsPathObj, ok := state.Parameters[argumentsPathField]
```
rgarcia, line 82: I think you can do the `.(string)` type cast here? I.e. go straight to 

```
			argumentsPath, ok := argumentsPathObj.(string)
```

executor/workflow_manager_sfn.go
ghirsch1: getting rid of this variable entirely

executor/workflow_manager_sfn.go
ghirsch1: getting rid of this variable entirely

ghirsch1:
Agreed, it ended up being very simple! Which is also good to know if we want to add some of the parallel states people have been asking about (though that'll probably require some more work in Hubble).

rgarcia:
Yeah it also makes me take a second look at some of the other integrations, e.g. the ECS one https://docs.aws.amazon.com/step-functions/latest/dg/connect-ecs.html. In theory this would allow for 1-1 aka "perfect" autoscaling since it would spin up ECS tasks in response to incoming work. The difficulty would be respecting maximum scale for a worker, which is currently handled by the fact that we run workers as ECS services with max counts (running snfcli to poll for available work).

ghirsch1:
We could potentially be clever with our use of placement constraints, but yeah it seems hard to properly enforce.

Also potentially this would be worse if we're spinning up a new task for each workflow. For the frequent workflows seems potentially better in many cases to leave a task running if we're going to need one in a second anyway.

rgarcia:
yeah good points. In Fargate placement constraints might not work...

One path that comes to mind is switching workflows to be the "scaled" entity instead of individual workers. Then workflow-manager could be responsible for ensuring that a maximum number of concurrent executions is respected before starting a new one, and the state machine can be free to do things like trigger ECS `StartTask`.

PR #271, opened 2020-11-11
Infrang 3783 support for parallel states

garrgordon:
## Link to JIRA:
https://clever.atlassian.net/browse/INFRANG-3783

## Overview:
Allows for creation of workflow definitions with parallel states and submission of those workflows on AWS step functions.  Since parallel states work by having their own state machines on each branch this required adding some recursive calls on state machine validation and expansion of resource values to ARN's.  Verified that this functions as expected and I am able to submit a workflow with a parallel state.

- [X] Update swagger.yml version
- [X] Run "make generate"

## Testing
Added test case for parsing parallel states.  Verified via manual ark submission using local workflow-manager that submitting workflows with parallel states works.

Parallel state workflow definition
Screen Shot 2020-11-13 at 2 55 53 PM
Map state workflow definition
Screen Shot 2020-11-13 at 2 57 40 PM

[Successful parallel state run](https://us-west-2.console.aws.amazon.com/states/home?region=us-west-2#/executions/details/arn:aws:states:us-west-2:589690932525:execution:clever-dev--sfncli-dummy-worker-INFRANG-3783-parallel--3--parallelStart:eb109017-38b5-4645-9a36-960b7182bb1f)

[Successful map state run](https://us-west-2.console.aws.amazon.com/states/home?region=us-west-2#/executions/details/arn:aws:states:us-west-2:589690932525:execution:clever-dev--sfncli-dummy-worker-INFRANG-3783-map--24--mapStart:fd82717f-01a7-4884-9d99-06967afcec22)

### DIFF
--- a/executor/workflow_manager_sfn.go
+++ b/executor/workflow_manager_sfn.go
@@ -66,11 +66,41 @@ func NewSFNWorkflowManager(sfnapi sfniface.SFNAPI, sqsapi sqsiface.SQSAPI, store
 // stateMachineWithFullActivityARNsAndParameters returns a new state machine.
 // Our workflow definitions contain state machine definitions with short-hand for resource names and parameter fields, e.g. "Resource": "name-of-worker"
 // Convert this shorthand into a new state machine with full details, e.g. "Resource": "arn:aws:states:us-west-2:589690932525:activity:production--name-of-worker"
-func stateMachineWithFullActivityARNsAndParameters(oldSM models.SLStateMachine, region, accountID, namespace string) *models.SLStateMachine {
+func stateMachineWithFullActivityARNsAndParameters(oldSM models.SLStateMachine, region, accountID, namespace string) (*models.SLStateMachine, error) {
 	sm := deepcopy.Copy(oldSM).(models.SLStateMachine)
+	var err error
 	for stateName, s := range sm.States {
 		state := deepcopy.Copy(s).(models.SLState)
-		if state.Type != models.SLStateTypeTask {
+		if state.Type == models.SLStateTypeParallel {
+			for index, branch := range state.Branches {
+				if branch == nil {
+					return nil, fmt.Errorf("Nil branch found in parallel state")
+				}
+				state.Branches[index], err = stateMachineWithFullActivityARNsAndParameters(*branch, region, accountID, namespace)
+				if err != nil {
+					return nil, err
+				}
+			}
+			sm.States[stateName] = state
+			continue
+		} else if state.Type == models.SLStateTypeMap {
+			if state.Iterator == nil {
+				return nil, fmt.Errorf("Nil iterator found in map state")
+			}
+			if state.Parameters == nil {
+				state.Parameters = map[string]interface{}{}
+			}
+			if _, copyingExecutionName := state.Parameters["_EXECUTION_NAME.$"]; !copyingExecutionName {
+				// Copies over of _EXECUTION_NAME so that its absence in input will not break things
+				state.Parameters["_EXECUTION_NAME.$"] = "$._EXECUTION_NAME"
+			}
+			state.Iterator, err = stateMachineWithFullActivityARNsAndParameters(*state.Iterator, region, accountID, namespace)
+			if err != nil {
+				return nil, err
+			}
+			sm.States[stateName] = state
+			continue
+		} else if state.Type != models.SLStateTypeTask {
 			continue
 		}
 		if strings.HasPrefix(state.Resource, "lambda:") {
@@ -87,7 +117,7 @@ func stateMachineWithFullActivityARNsAndParameters(oldSM models.SLStateMachine, 
 		}
 		sm.States[stateName] = state
 	}
-	return &sm
+	return &sm, nil
 }
 
 // stateMachineWithDefaultRetriers creates a new state machine that has the following retry properties on every state's retry array:
@@ -151,7 +181,10 @@ func (wm *SFNWorkflowManager) describeOrCreateStateMachine(ctx context.Context, 
 	}
 
 	// state machine doesn't exist, create it
-	awsStateMachine := stateMachineWithFullActivityARNsAndParameters(*wd.StateMachine, wm.region, wm.accountID, namespace)
+	awsStateMachine, err := stateMachineWithFullActivityARNsAndParameters(*wd.StateMachine, wm.region, wm.accountID, namespace)
+	if err != nil {
+		return nil, err
+	}
 	awsStateMachine = stateMachineWithDefaultRetriers(*awsStateMachine)
 	awsStateMachineDefBytes, err := json.MarshalIndent(awsStateMachine, "", "  ")
 	if err != nil {
--- a/executor/workflow_manager_sfn_test.go
+++ b/executor/workflow_manager_sfn_test.go
@@ -165,12 +165,49 @@ var stateMachineWithFullActivityARNsAndParametersTests = []stateMachineWithFullA
 			},
 		},
 	},
+	stateMachineWithFullActivityARNsAndParametersTest{
+		name: "map state adds EXECUTION_NAME and expands iterator states",
+		input: models.SLStateMachine{
+			States: map[string]models.SLState{
+				"foostatemap": models.SLState{
+					Type: models.SLStateTypeMap,
+					Iterator: &models.SLStateMachine{
+						StartAt: "foostate",
+						States: map[string]models.SLState{
+							"foostate": models.SLState{
+								Type:     models.SLStateTypeTask,
+								Resource: "resource-name",
+							},
+						},
+					},
+				},
+			},
+		},
+		wantSMStates: map[string]models.SLState{
+			"foostatemap": models.SLState{
+				Type: models.SLStateTypeMap,
+				Parameters: map[string]interface{}{
+					"_EXECUTION_NAME.$": "$._EXECUTION_NAME",
+				},
+				Iterator: &models.SLStateMachine{
+					StartAt: "foostate",
+					States: map[string]models.SLState{
+						"foostate": models.SLState{
+							Type:     models.SLStateTypeTask,
+							Resource: "arn:aws:states:region:accountID:activity:namespace--resource-name",
+						},
+					},
+				},
+			},
+		},
+	},
 }
 
 func TestStateMachineWithFullActivityARNsAndParameters(t *testing.T) {
 	for _, test := range stateMachineWithFullActivityARNsAndParametersTests {
-		smWithFullActivityARNs := stateMachineWithFullActivityARNsAndParameters(test.input, "region", "accountID", "namespace")
+		smWithFullActivityARNs, err := stateMachineWithFullActivityARNsAndParameters(test.input, "region", "accountID", "namespace")
 		assert.Equal(t, test.wantSMStates, smWithFullActivityARNs.States)
+		assert.NoError(t, err)
 	}
 }
 
--- a/handler.go
+++ b/handler.go
@@ -33,14 +33,118 @@ func (h Handler) HealthCheck(ctx context.Context) error {
 	return nil
 }
 
+func validateStateMachine(sm *models.SLStateMachine) error {
+	if len(sm.States) == 0 {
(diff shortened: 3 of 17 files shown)


ghirsch1:
I haven't looked at the code, but could you add to the PR links to the workflows you used to test with? E.g. the DB entry, the step functions page, etc.?

Also would it be easy to add the same functionality for the Map state mentioned on the ticket?

resources/workflow_definitions.go
```diff
@@ -53,57 +53,66 @@ func stateExists(stateName string, stateMachine *models.SLStateMachine) bool {
...
-				return true
+				return true, nil
 			}
 		case models.SLStateTypeSucceed, models.SLStateTypeFail:
 			// these states don't contain transitions
-		case models.SLStateTypeParallel:
-			panic("parallel states not supported yet")
 		default:
-			panic(fmt.Sprintf("%s states not supported yet", state.Type))
+			return false, fmt.Errorf("%s states not supported yet", state.Type)
```
tnsardesai, line 92: Nice catch changing this from panic to an error

tnsardesai (approved):
LGTM but wait on 1 more review. You can do `roll a d2` and assign it to either gavi or taylor

garrgordon:
> I haven't looked at the code, but could you add to the PR links to the workflows you used to test with? E.g. the DB entry, the step functions page, etc.?
> 
> Also would it be easy to add the same functionality for the Map state mentioned on the ticket?

sfncli-dummy-worker:INFRANG-3783-parallel2 is the workflow-definitions dev table https://us-west-1.console.aws.amazon.com/dynamodb/home?region=us-west-1#tables:selected=workflow-manager-dev-v3-workflow-definitions;tab=items

https://us-west-2.console.aws.amazon.com/states/home?region=us-west-2#/statemachines/view/arn:aws:states:us-west-2:589690932525:stateMachine:clever-dev--sfncli-dummy-worker-INFRANG-3783-parallel--2--parallelStart

I'll try to work on adding support for the map state as well

ghirsch1:
> I'll try to work on adding support for the map state as well

If it is nontrivial and you'd have to choose between that and the Hubble piece, feel free to do whichever is most interesting to you.

garrgordon:
> > I'll try to work on adding support for the map state as well
> 
> If it is nontrivial and you'd have to choose between that and the Hubble piece, feel free to do whichever is most interesting to you.

Added support for it after our discussion yesterday I decided on a solution, but if you have any alternative ideas I'm open to it.  I'm doing some validation now in the handler that checks for a starting map state if it's a map state as a start state we check that the _EXECUTION_NAME variable is copied to the inputs using the Parameters field.  If there is a map state further down the line the the jobs themselves can be responsible for making sure the input with _EXECUTION_NAME is passed correctly

ghirsch1:
> If there is a map state further down the line the jobs themselves can be responsible for making sure the input with _EXECUTION_NAME is passed correctly

I think I'd prefer to just have it always get copied so that engineers don't have to worry about including it in this special case (unless there's an advantage to having engineers do it that I'm not seeing).

> Could you add to the PR links to the workflows you used to test with?

Links, screenshots, etc. are always super helpful to have (this time for the Map tests you're doing). Also while it works to put them in comments, it's most convenient imo to put them on the PR description so that they're easy to find.

executor/workflow_manager_sfn.go
```diff
@@ -70,7 +70,17 @@ func stateMachineWithFullActivityARNsAndParameters(oldSM models.SLStateMachine,
...
 		state := deepcopy.Copy(s).(models.SLState)
-		if state.Type != models.SLStateTypeTask {
+		if state.Type == models.SLStateTypeParallel {
+			for index, branch := range state.Branches {
+				state.Branches[index] = stateMachineWithFullActivityARNsAndParameters(*branch, region, accountID, namespace)
+			}
+			sm.States[stateName] = state
+			continue
+		} else if state.Type == models.SLStateTypeMap {
+			state.Iterator = stateMachineWithFullActivityARNsAndParameters(*state.Iterator, region, accountID, namespace)
```
ghirsch1, line 80: In general when dereferencing pointers it's nice to do a nil check so that if somehow they're nil we can error rather than panicking.

swagger.yml
```diff
@@ -678,12 +678,19 @@ definitions:
...
+        type: string
       OutputPath:
         type: string
       ResultPath:
         type: string
       Result:
         type: string
+      ResultSelector:
+        additionalProperties:
+          type: string
```
ghirsch1, line 691: Curious what this looks like in practice. Presumably it's for merging the outputs together?

resources/kitchensink.go
```diff
@@ -31,14 +31,43 @@ func KitchenSinkWorkflowDefinition(t *testing.T) *models.WorkflowDefinition {
 				},
 				"second-state": models.SLState{
 					Type:     models.SLStateTypeTask,
-					Next:     "end-state",
+					Next:     "parallel-state",
 					Resource: "fake-resource-2",
 					End:      false,
 					Retry:    []*models.SLRetrier{},
 				},
+				"parallel-state": models.SLState{
```
ghirsch1, line 39: Nice, thanks for adding this. Wouldn't hurt to add a map example too

handler.go
```diff
@@ -33,6 +33,25 @@ func (h Handler) HealthCheck(ctx context.Context) error {
 	return nil
 }
 
+func validateStartingMapState(sm *models.SLStateMachine) error {
```
ghirsch1, line 36: Why just for the starting state? Doesn't this need to be true any time a map state shows up?

ghirsch1 (commented):
Nice, this looks great! Left a few comments and would love to see some examples of this working in practice

executor/workflow_manager_sfn.go
garrgordon: d7cba28675f3eb2c1d1d543106568c4bc2dd2db4

resources/kitchensink.go
garrgordon: d7cba28675f3eb2c1d1d543106568c4bc2dd2db4

handler.go
garrgordon: Yep good point fixed that here aabb16067b1259eb312de89ee296bf45f17d3610

swagger.yml
garrgordon: It allows you to merge or filter out various elements of your output https://docs.aws.amazon.com/step-functions/latest/dg/input-output-inputpath-params.html

garrgordon:
> I think I'd prefer to just have it always get copied so that engineers don't have to worry about including it in this special case (unless there's an advantage to having engineers do it that I'm not seeing).

So  guess I was concerned that engineers might get confused if the value magically appeared and we store a workflow definition that's different from what they actually put in ark-config.  but I can see how it might be more convenient as well.  Either way we can validate to make sure it's always present aabb16067b1259eb312de89ee296bf45f17d3610
Or check to see if they've added it and insert it if they haven't 73ed2088da470cb6358328eb0734cdfe2fc41f73
Whichever you think is best

swagger.yml
garrgordon: Yep let's you merge or filter output. https://docs.aws.amazon.com/step-functions/latest/dg/input-output-inputpath-params.html  Added some examples to the PR details

resources/kitchensink.go
garrgordon: d7cba28675f3eb2c1d1d543106568c4bc2dd2db4

handler.go
garrgordon: Yep good point fixed it here aabb16067b1259eb312de89ee296bf45f17d3610

garrgordon:
> Links, screenshots, etc. are always super helpful to have (this time for the Map tests you're doing). Also while it works to put them in comments, it's most convenient imo to put them on the PR description so that they're easy to find.
Added screenshots of workflow definitions for parallel and map states and links to successful runs to the PR description.  They're using the ResultSelector field since you mentioned you were curious about that.

executor/workflow_manager_sfn_test.go
```diff
@@ -169,7 +169,7 @@ var stateMachineWithFullActivityARNsAndParametersTests = []stateMachineWithFullA
 
 func TestStateMachineWithFullActivityARNsAndParameters(t *testing.T) {
 	for _, test := range stateMachineWithFullActivityARNsAndParametersTests {
-		smWithFullActivityARNs := stateMachineWithFullActivityARNsAndParameters(test.input, "region", "accountID", "namespace")
+		smWithFullActivityARNs, _ := stateMachineWithFullActivityARNsAndParameters(test.input, "region", "accountID", "namespace")
 		assert.Equal(t, test.wantSMStates, smWithFullActivityARNs.States)
```
ghirsch1, line 173:
```suggestion
		smWithFullActivityARNs, err := stateMachineWithFullActivityARNsAndParameters(test.input, "region", "accountID", "namespace")
		assert.Equal(t, test.wantSMStates, smWithFullActivityARNs.States)
		assert.NoError(t, err)
```

handler.go
```diff
@@ -33,16 +33,128 @@ func (h Handler) HealthCheck(ctx context.Context) error {
...
+		return fmt.Errorf("Map state must use the ItemsPath field")
+	}
+	if state.Parameters == nil {
+		return fmt.Errorf("Map state must use the Paramters field in order to copy _EXECUTION_NAME to inputs")
+	}
+
+	if _, copyingExecutionName := state.Parameters["_EXECUTION_NAME.$"]; !copyingExecutionName {
+		// Copies over of _EXECUTION_NAME so that its absence in input will not break things
+		state.Parameters["_EXECUTION_NAME.$"] = "$._EXECUTION_NAME"
+	}
```
ghirsch1, line 144: This isn't validation, it's doing a modification (the first part technically is, but it's only a requirement due to the second part). Better to put it here (see how the glue type does something similar with parameters) https://github.com/Clever/workflow-manager/blob/master/executor/workflow_manager_sfn.go#L69

handler.go
```diff
@@ -33,16 +33,128 @@ func (h Handler) HealthCheck(ctx context.Context) error {
 	return nil
 }
 
+func validateStateMachine(sm *models.SLStateMachine) error {
```
ghirsch1, line 36: 💯 Wow, awesome to have all this validation now! Thanks for adding it!

ghirsch1:
> Also, the engineer will already have to add the Parameters field on their own because we can't know what the correct field name is to reference their input array...

Technically I don't think specifying other parameters is required since in theory you could have a map state that splits based on the number of elements in the array but doesn't care about the values. But yeah, in practice it should generally almost always be specified.

But given the following I think it's better to add in the shadows
1. Engineers will never want to set a different value for it
2. It is generally required, and will break things if not present

ghirsch1:
Also thanks for adding the details and links to the workflow details on the PR! One small thing for you to personally be aware of when writing workflows, in general you probably don't want to set `ResultPath: $` since it wipes the previous state. Instead, typically better to set `ResultPath: $.WHATEVER_NAME_YOU_WANT`.

Once this is ready we should add some details to this wiki page about parallel and map, and we can put this note about ResultPath there too https://clever.atlassian.net/wiki/spaces/ENG/pages/246579345/Workflows

handler.go
garrgordon: b7955e71618b1a977f5f5a77c2dc4cf953f9cadf

garrgordon:
> Also thanks for adding the details and links to the workflow details on the PR! One small thing for you to personally be aware of when writing workflows, in general you probably don't want to set `ResultPath: $` since it wipes the previous state. Instead, typically better to set `ResultPath: $.WHATEVER_NAME_YOU_WANT`.
> 
> Once this is ready we should add some details to this wiki page about parallel and map, and we can put this note about ResultPath there too https://clever.atlassian.net/wiki/spaces/ENG/pages/246579345/Workflows

Ah I'll keep that in mind.  Yeah I'll work on documentation once I've merged

ghirsch1 (approved):
🚀 :shipit:

Sayan-:
yoooooooooooo this is rad

ghirsch1:
eyyyyy glad to see you're still following the Clever repos 😄

PR #301, opened 2021-07-09
podOnly

tnsardesai:
pod migration

Tested deployment to pods in my dev environment so directly deploying to podOnly

After merge:
- ark start --upstreams
- monitor


taylor-sutton (approved):
Also update https://github.com/Clever/ark/blob/master/config/load.go#L36-L39

rgarcia (approved):

PR #317, opened 2022-08-19
turn on state machine logging in dev

rgarcia:
## Link to JIRA:

## Overview:

Turns on step functions execution logging in non-production envs so we can start to understand these logs: https://docs.aws.amazon.com/step-functions/latest/dg/cw-logs.html


ChrisScotMartin:
Looks like it's still failing `make test` due to the changes in executor. Might want to fix those tests first.

rgarcia:
oops, just pushed a fix that should fix the tests

ChrisScotMartin (approved):
LGTM!

PR #319, opened 2022-08-30
factor out some update loop logic into separate package

rgarcia:
## Link to JIRA:

## Overview:
The most involved section of the update loop is a section that pulls in the full execution history whenever a workflow fails and populates a "lastJob" field on the workflow object with more details about the failure. This PR factors out the core piece of that, a function named `updateWorkflowLastJob`. This will assist in using the logic elsewhere (forthcoming lambda fn).

- [ ] Run "make generate"

### DIFF
--- a/wfupdater/lastjob.go
+++ b/wfupdater/lastjob.go
@@ -1,0 +1,43 @@
+package wfupdater
+
+import (
+	"context"
+
+	"github.com/Clever/workflow-manager/gen-go/models"
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/service/sfn"
+	"github.com/aws/aws-sdk-go/service/sfn/sfniface"
+	"gopkg.in/Clever/kayvee-go.v6/logger"
+)
+
+// UpdateWorkflowLastJob queries AWS sfn for the latest events of the given state machine,
+// and it uses these events to populate LastJob within the WorkflowSummary.
+// This method is heavily inspired by UpdateWorkflowHistory, however updateWorkflowLastJob only
+// finds the last Job which is all that is needed to determine the state where a workflow has failed.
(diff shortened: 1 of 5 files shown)


wfupdater/lastjob.go
```diff
@@ -0,0 +1,43 @@
...
+)
+
+// UpdateWorkflowLastJob queries AWS sfn for the latest events of the given state machine,
+// and it uses these events to populate LastJob within the WorkflowSummary.
+// This method is heavily inspired by UpdateWorkflowHistory, however updateWorkflowLastJob only
+// finds the last Job which is all that is needed to determine the state where a workflow has failed.
+// We try to use the sfn GetExecutionHistory API endpoint somewhat sparingly because
+// it has a relatively low rate limit, and it can take multiple calls per workflow
+// to retrieve the full event history for some workflows.
+func UpdateWorkflowLastJob(ctx context.Context, sfnapi sfniface.SFNAPI, execARN string, workflow *models.Workflow) error {
```
rgarcia, line 20: this function was copy-pasted and then changed to take in two additional arguments (`sfnapi sfniface.SFNAPI, execARN string`) in order to detach it from the struct state it previously relied on

wfupdater/processevents.go
```diff
@@ -0,0 +1,291 @@
+package wfupdater
```
rgarcia, line 1: this entire file is copy-pasted. Only thing changed was `processEvents` -> `ProcessEvents`

rgarcia:
FYI @ChrisScotMartin the codebuild failure is not meaningful, it just means that someone (me) introduced an ark db config file to this repo and has yet to merge it to master

ChrisScotMartin (approved):
Thanks for the comments about copy/pasting. 

LGTM

PR #321, opened 2022-09-01
introduce kinesis consumer for (future) processing of SFN execution events

rgarcia:

## Overview

Best to review this commit by commit:

1. Add a kinesis stream ark db config. This stream will eventually be subscribed to log groups that SFN2. 
2. Add a lambda app to the repo that does nothing. (Also ran `make generate` so there is some random gen-go changes due to using latest wag) 
3. Implement the lambda function. This is the main part of this PR

## Testing
I'd like to get this merged to master so that I can deploy and test the lambda in a dev environment (cannot do this w/o a published master build)

## Rollout
Deploying this should have no functional change since the kinesis streams are not receiving any data.


cmd/sfn-execution-events-consumer/main.go
```diff
@@ -1,44 +1,268 @@
...
+			return fmt.Errorf("unexpected no output for succeeded event. ARN %s", evt.ExecutionARN)
+		} else {
+			update.Output = evt.Details.Output
+		}
+	}
+
+	// we consider any successful or or canceled workflow as "resolved"
+	if update.Status != nil && (*update.Status == models.WorkflowStatusSucceeded || *update.Status == models.WorkflowStatusCancelled) {
+		update.ResolvedByUser = aws.Bool(true)
+	}
```
rgarcia, line 196: the logic above this line is the (eventual) replacement of the logic in the update loop found here: https://github.com/Clever/workflow-manager/blob/d062148c8f91ce9830d33f3c6d9827b4d60dfa43/executor/workflow_manager_sfn.go#L514-L524

also see "Translating Execution History Events To a Workflow Status" in the [spec](https://docs.google.com/document/d/1JzdQr_h8wwdbiu3kRc_9G0CiHtQOEyWPuKtpsdqWlVc/edit#heading=h.7vhb5sq732tj)

cmd/sfn-execution-events-consumer/main.go
```diff
@@ -1,44 +1,268 @@
...
+			}
+		}
+		logger.FromContext(ctx).CounterD("workflow-failed", 1, logger.M{
+			"workflow-name":       workflow.WorkflowDefinition.Name,
+			"workflow-version":    workflow.WorkflowDefinition.Version,
+			"workflow-id":         workflow.ID,
+			"failed-job-name":     failedJob,
+			"failed-job-resource": failedJobResource,
+		})
+	}
```
rgarcia, line 223: this `if` block is a replacement for the logic in the update found here: https://github.com/Clever/workflow-manager/blob/d062148c8f91ce9830d33f3c6d9827b4d60dfa43/executor/workflow_manager_sfn.go#L525-L546

launch/sfn-execution-events-stream.yml
```diff
@@ -0,0 +1,38 @@
...
+    severity: minor
+    parameters:
+      threshold: .75
+      period: 60
+      evaluationPeriods: 5
+  # consumers falling behind more than 30 minutes
+  - type: KinesisDelaysAlarm
+    severity: minor
+    runbookURL: ""
+    tips: "Workflow Manager is falling behind state machine events, consider increasing shard count of the kinesis stream, which will create more lambda consumers."
```
nbhatia823, line 25: is increasing the shard count a simple step? maybe in the future a runbook could be helpful

cmd/sfn-execution-events-consumer/main.go
```diff
@@ -1,44 +1,268 @@
...
+	case "ExecutionSucceeded":
+		update.Status = ptrStatus(models.WorkflowStatusSucceeded)
+		if evt.Details.Output == nil {
+			return fmt.Errorf("unexpected no output for succeeded event. ARN %s", evt.ExecutionARN)
+		} else {
+			update.Output = evt.Details.Output
+		}
+	}
+
+	// we consider any successful or or canceled workflow as "resolved"
```
nbhatia823, line 193: nit: or or

launch/sfn-execution-events-consumer.yml
```diff
@@ -0,0 +1,34 @@
+run:
+  type: lambda
+  architecture: arm64
```
nbhatia823, line 3: need this to be `x86_64` since in uw1? Or could it run in uw2? If x86 reminder to use 0.2.4 `lambda.mk`

nbhatia823 (approved):

launch/sfn-execution-events-consumer.yml
rgarcia: ahh good catch. Has to be in us-west-1 where the kinesis stream is... although now that I think about it we should probably move the kinesis stream to us-west-2 since the logs will be coming from step functions (which we run in us-west-2)

launch/sfn-execution-events-stream.yml
rgarcia: yeah it's as simple as updating the `shardCount: 1` config line above and `ark db start ...`

PR #324, opened 2022-09-13
scale up kinesis stream in prod

rgarcia:
starting to see the consumer fall behind as i turn on event log processing for more workflows


ChrisScotMartin (approved):
I'm guessing this will help with some of the kinesis alerts we've been getting.

PR #325, opened 2022-09-14
resolved by user should be set for failed workflows

rgarcia:
## Link to JIRA:
[Link to JIRA](https://clever.atlassian.net/browse/INFRANG-4473)

## Overview:
Fixing a bug where failed workflows aren't showing up in hubble, which searches for resolvedByUser: true.

I'm actually not sure how it used to get set to true for failed workflows. The code changed in this PR diverges from the analogous code in the update loop: https://github.com/Clever/workflow-manager/blob/4cdb21d053b3f2a2035256214bb76c36bfea84d9/executor/workflow_manager_sfn.go#L576-L578

- [ ] Run "make generate"


andruwm (approved):
lgtm

PR #328, opened 2022-09-20
Revert "resolved by user should be set for failed workflows"

rgarcia:
Reverts Clever/workflow-manager#325

The actual problem was event ordering and was solved by https://github.com/Clever/workflow-manager/pull/327


rgarcia:
cc @meliaj I think this is why you're not able to re-run failed workflows in hubble

taylor-sutton (approved):

PR #329, opened 2022-09-22
remove update loop code

rgarcia:
## Link to JIRA:
[Link to JIRA](https://clever.atlassian.net/browse/INFRANG-4473)

## Overview:
Removes all SQS and LaunchDarkly related code.

- [ ] Run "make generate"


executor/log_helpers.go
```diff
@@ -1,53 +1,12 @@
...
-	})
-}
-
-func logPendingWorkflowUpdateLag(wf models.Workflow) {
-	log.TraceD("pending-workflow-update-lag", logger.M{
-		"id":                      wf.ID,
-		"update-loop-lag-seconds": int(time.Now().Sub(time.Time(wf.LastUpdated)) / time.Second),
-	})
-}
-
```
ChrisScotMartin, line 50: So no more logging because we are pulling from events and have visibility with Hubble?

And update-loop-lag-seconds has no meaning anymore since we don't have an update-loop :)

ChrisScotMartin (approved):
Nice! That's a lot of cleanup! It's good to see significantly reduced overall complexity.

executor/log_helpers.go
rgarcia: there is still lag but we have alerts on it via the kinesis stream where state machine events are delivered and consumed https://github.com/Clever/workflow-manager/blob/master/launch/sfn-execution-events-stream.yml#L14-L38

PR #330, opened 2022-10-04
Add better context to error messages relating to terminal states

andruwm:
## Overview:
There are a bunch of errors causing workflows not to be updated and they are all related to this error. This PR adds additional context so the problem can be debugged more easily.

### DIFF
--- a/cmd/sfn-execution-events-consumer/main.go
+++ b/cmd/sfn-execution-events-consumer/main.go
@@ -6,6 +6,7 @@ import (
 	"context"
 	"encoding/base64"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"io/ioutil"
 	"log"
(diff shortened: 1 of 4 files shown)


cmd/sfn-execution-events-consumer/main.go
```diff
@@ -258,8 +259,8 @@ func (h Handler) handleHistoryEvent(ctx context.Context, evt HistoryEvent) error
 	}
 	logger.FromContext(ctx).InfoD("update-workflow", logger.M(update.Map()))
 	if err := h.store.UpdateWorkflowAttributes(ctx, execID, update); err != nil {
-		if err == store.ErrUpdatingWorkflowFromTerminalToNonTerminalState {
-			logger.FromContext(ctx).Info("workflow-already-terminal")
+		if errors.Is(err, store.ErrUpdatingWorkflowFromTerminalToNonTerminalState) {
```
taylor-sutton, line 262: It looks like these errors in DataDog don't have the workflow ID/execution ID. There might be something fishy with https://github.com/Clever/workflow-manager/blob/29259e2391e9eadc2845a6acce347e15657e5790/cmd/sfn-execution-events-consumer/main.go#L184-L189. But can be handled in separate PR if you want that info.

taylor-sutton (approved):

cmd/sfn-execution-events-consumer/main.go
andruwm: Removed the defers for now to see if that helps. If it makes it worse I'll address in a followup PR.

PR #332, opened 2022-10-05
INFRANG-5061 - Fix case handling workflow running state

andruwm:
## Link to JIRA:
[INFRANG-5061](https://clever.atlassian.net/browse/INFRANG-5061)

## Overview:
This PR adds a script to force updates on workflows in a stuck state. Stuck state may occur when the SFN state pipeline has bugs and loses data. 

Additionally, I have added notes about the special case handling of event ID 2.

### DIFF
--- a/cmd/sfn-execution-events-consumer/main.go
+++ b/cmd/sfn-execution-events-consumer/main.go
@@ -187,8 +187,13 @@ func (h Handler) handleHistoryEvent(ctx context.Context, evt HistoryEvent) error
 	logger.FromContext(ctx).AddContext("execution-id", execID)
 	logger.FromContext(ctx).AddContext("state-machine-name", smName)
 	logger.FromContext(ctx).AddContext("aws-event-type", evt.Type)
-
 	var update store.UpdateWorkflowAttributesInput
+	if evt.ID == "2" {
+		update.Status = ptrStatus(models.WorkflowStatusRunning)
+		logger.FromContext(ctx).InfoD("update-workflow", logger.M(update.Map()))
+		return swallowOutOfOrderStateError(ctx, h.store.UpdateWorkflowAttributes(ctx, execID, update))
+	}
+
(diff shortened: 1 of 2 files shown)


taylor-sutton:
> The existing logic makes the assumption that step function event indexes are 1:1 with the index for the given execution in cloud watch logs. As shown in the attached pictures this is not actually the case. Event indexes appear to be shared by all executions within a log stream. 

I'm not sure I follow. Looking at https://us-west-2.console.aws.amazon.com/cloudwatch/home?region=us-west-2#logsV2:log-groups/log-group/$252Faws$252Fvendedlogs$252Fstates$252Fproduction--data-pipeline-master--27--start/log-events/states$252Fproduction--data-pipeline-master--27--start$252F2022-10-05-23$252Fc837649c for example, it seems like there are multiple executions going on in this stream, but each execution has its own indexes starting from 1.

I wonder if maybe there was an idea to not repeatedly issue updates to a workflow from Running back to Running? But the existing code that returns quickly on event ID 2 doesn't achieve that...

taylor-sutton:
Self quote - 

> I wonder if maybe there was an idea to not repeatedly issue updates to a workflow from Running back to Running? But the existing code that returns quickly on event ID 2 doesn't achieve that...

I think maybe that's it. Notice how the existing code never tries to do a status update when it sees TaskStateEntered  _except_ on ID 2

andruwm:
Ah you know what, I think you're right. I was conflating the `_source` field with the execution. I guess there can be multiple sources within an execution, each being a separate worker/lambda within a workflow? Still learning the workflow architecture here. 

Given these new points, I'm not sure how we can idempotently set updates for running state when the first running state index isn't consistent. We don't have any context of the current state until we call out to DynamoDB. I'll have to think on this some, but open to ideas.

taylor-sutton (approved):

PR #338, opened 2022-12-05
Fix data loss

andruwm:
## Link to JIRA:
[INFRANG-5061](https://clever.atlassian.net/browse/INFRANG-5061)

## Overview:
This PR fixes 2 bugs. 
1. Handle the case where event ID number 2 may be a terminal event. I don't know for sure that this case exists but it's an easy change and doesn't make the code flow more messy.
2. Don't fail and throw out all unprocessed data on any error. I haven't dug in, but I suspect this may be ay least one source of our data loss. Given that this change was easy I wanted to try it out and see if I can save time on really digging into why this is occurring.

These aren't random shots in the dark, more like educated guesses. My hope is this 20min spent throwing together this PR fixing at least one bug that certainly exists will save me a few hours of debugging.

### DIFF
--- a/cmd/sfn-execution-events-consumer/main.go
+++ b/cmd/sfn-execution-events-consumer/main.go
@@ -27,6 +27,7 @@ import (
 	"github.com/aws/aws-sdk-go/service/sfn"
 	"github.com/aws/aws-sdk-go/service/sfn/sfniface"
 	"github.com/go-openapi/strfmt"
+	"github.com/hashicorp/go-multierror"
 
 	"github.com/Clever/workflow-manager/gen-go/models"
 	dynamodbgen "github.com/Clever/workflow-manager/gen-go/server/db/dynamodb"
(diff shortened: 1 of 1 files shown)


ChrisScotMartin (approved):
🤞

cmd/sfn-execution-events-consumer/main.go
```diff
@@ -27,6 +27,7 @@ import (
 	"github.com/aws/aws-sdk-go/service/sfn"
 	"github.com/aws/aws-sdk-go/service/sfn/sfniface"
 	"github.com/go-openapi/strfmt"
+	"github.com/hashicorp/go-multierror"
```
ChrisScotMartin, line 30: That's a new one to me. Any significant advantage over just a []errors and then checking if the len is 0? I've seen that approach used in our code elsewhere. 

The append is shorter at least.

andruwm: This package was already in use in this repo otherwise I may have opted to just use `[]error`. The issue with error slices is that they are clunky. Clunky to add to, clunky to combine into a single useful error, clunky to pass around etc. Many times in functions where you need to combine errors, not all code paths actually need to so you end up returning a single error wrapped in a slice. This package handles all of the hard parts for you, and will even flatten multiple single errors that actually contain a multi error under the hood. The process becomes only a little less ergonomic than a regular error.

PR #344, opened 2023-02-23
Upgrade to wagv9.1

nbhatia823:
PR created via Microplane

## What does this PR do?

This PR is a follow up to the wag v9 rollout. It is our cleanup PR that has a few purposes:

1. Updates wag to 9.1.0 where we have introduced a new client creation API to make things simpler for eng. From now on, we can use a helper library called `clientconfig` to create wag clients
2. Updates the golang.mk, wag.mk, and launch-gen
3. Replaces all use of kayvee 6 with our new standard kayvee 7
4. Improves `make generate` by always copying over updated generated client/models to vendor

If you have questions, please ask here or in #oncall-infra.


ChrisScotMartin (approved):
Sweet, looks good to me!

PR #351, opened 2023-07-21
log when zombie execution is found

tnsardesai:
## Link to JIRA:
[INFRANG-4733](https://clever.atlassian.net/browse/INFRANG-4733)

## Overview:
Zombie executions are defined as workflows that exist in DDB but don't have a step function execution. This can happen if workflow was added to DDB but starting execution failed and the workflow was never removed from ddb. I looked over the code and I don't see any spot where we do this. I also investigated our logs and I didn't see any evidence of zombie workflows so we must have fixed the underlying issue already. Lets add this log line to track zombie executions and get alerted on it.

In hubble we are filtering out real workflows thinking they are zombies based on https://github.com/Clever/hubble/pull/96/files. Instead of a query like this it is better to just eliminate zombie workflows which is why this log is added. In the mean time lets revert the hubble PR so that real workflows are not filtered out.

- [na] Update swagger.yml version
- [na] Run "make generate"


tnsardesai:
Added a comment on the ticket in the alert on what to do when it is fired

jakegut (approved):

PR #358, opened 2023-08-14
Oncall fix 500 spam get workflow

tnsardesai:
## Link to JIRA:
oncall

## Overview:
https://app.datadoghq.com/logs?query=service%3Aworkflow-manager%20env%3Aproduction%20status%3Aerror%20%40status-code%3A500%20&agg_q=%40path&analyticsOptions=%5B%22bars%22%2C%22dog_classic%22%2Cnull%2Cnull%5D&cols=host%2Cservice&index=&messageDisplay=inline&stream_sort=desc&viz=query_table&from_ts=1691877731318&to_ts=1692050531318&live=true

we are seeing a spam of 500s in workflow-manager

Recently we updated hubble to remove a query which was filtering out zombie workflows https://clever.atlassian.net/browse/INFRANG-4733. Looks like now we have some zombie workflows (I haven't investigated why yet) which are now getting queried by hubble leading to this increase in 500s. We should fix the issue that caused these zombie workflows so that the they get deleted from dynamodb but in the meantime this error shouldn't be treated as a 500. Instead it should just be ignored so that `getWorkflowbyId` returns the workflow stored in dynamodb which shows that it failed.

- [ ] Run "make generate"

### DIFF
--- a/executor/workflow_manager_sfn.go
+++ b/executor/workflow_manager_sfn.go
@@ -567,13 +567,6 @@ func (wm *SFNWorkflowManager) UpdateWorkflowSummary(ctx context.Context, workflo
 }
 
 func (wm *SFNWorkflowManager) UpdateWorkflowHistory(ctx context.Context, workflow *models.Workflow) error {
-	// Avoid the extraneous processing for executions that have already stopped.
-	// This also prevents the WM "cancelled" state from getting overwritten for workflows cancelled
-	// by the user after a failure.
-	if resources.WorkflowIsDone(workflow) {
-		return nil
-	}
-
 	// Pull in execution history to populate jobs array
 	// Each Job corresponds to a type={Task,Choice,Succeed} state, i.e. States we have currently tested and supported completely
 	// We only create a Job object if the State has been entered.
--- a/executor/workflow_manager_sfn.go
+++ b/executor/workflow_manager_sfn.go
@@ -567,6 +567,13 @@ func (wm *SFNWorkflowManager) UpdateWorkflowSummary(ctx context.Context, workflo
 }
 
 func (wm *SFNWorkflowManager) UpdateWorkflowHistory(ctx context.Context, workflow *models.Workflow) error {
+	// Avoid the extraneous processing for executions that have already stopped.
+	// This also prevents the WM "cancelled" state from getting overwritten for workflows cancelled
+	// by the user after a failure.
+	if resources.WorkflowIsDone(workflow) {
+		return nil
+	}
+
 	// Pull in execution history to populate jobs array
 	// Each Job corresponds to a type={Task,Choice,Succeed} state, i.e. States we have currently tested and supported completely
 	// We only create a Job object if the State has been entered.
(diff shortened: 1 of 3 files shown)


executor/workflow_manager_sfn.go
```diff
@@ -567,6 +567,13 @@ func (wm *SFNWorkflowManager) UpdateWorkflowSummary(ctx context.Context, workflo
 }
 
 func (wm *SFNWorkflowManager) UpdateWorkflowHistory(ctx context.Context, workflow *models.Workflow) error {
+	// Avoid the extraneous processing for executions that have already stopped.
```
taylor-sutton, line 570: Is this safe? This seems like a somewhat significant behavior change: it means we will never update `Jobs` for finished workflows. Like: `Jobs` is _only_ updated by this function (outside of special cases), and looking at https://github.com/Clever/workflow-manager/blob/9f58d4ebd698c16932e197be17909dd60d3a5b93/resources/workflows.go#L68, if `Jobs` is empty, this function will short circuit and `Jobs` will stay empty.

tnsardesai: Actually just adding the conditional should fix the spam, this is just an optimization I added. To make it safe in handler.go in `GetWorkflowByID` I updated the order of the functions so that jobs list is updated before we update the summary.

So if job list is updated before the summary then if workflow summary is set to done we shouldn't expect any changes to the jobs or any new jobs. Thats why I think the overall change should be safe.

tnsardesai: Actually looking at https://github.com/Clever/workflow-manager/blob/9f58d4ebd698c16932e197be17909dd60d3a5b93/handler.go#L491-L507. we also call this function when retrying a workflow which means this optimization will break retrying! I can remove this but the rest of the change should be fine (including the change in order for updateworkflowhistory and updateworkflowsummary)

taylor-sutton (approved):

executor/workflow_manager_sfn.go
taylor-sutton: I think this is broken even outside of calling Retry. For example in this flow:
1. Workflow is started
2. Workflow runs to completion and succeeds without anyone calling `GetWorkflowByID`. At this point, `Jobs` is empty and `status` is success because of the async `sfn-execution-events-consumer`
3. Someone calls `GetWorkflowByID` on this workflow. The `UpdateHistory` part short circuits immediately and `Jobs` is never updated.

You can refer to https://clever.atlassian.net/wiki/spaces/~620990898/pages/3019833367/What+s+in+a+workflow or the code for

tnsardesai: TIL how this works `sfn-execution-events-consumer`. I will take a look at that doc as I investigate zombie workflows this week

tnsardesai: I didn't realize that we don't update dynamo db until someone actually queries `GetWorkflowByID`

taylor-sutton: Yeah no problem... this system is complicated, to say the least, it just happens to be top of mind due to last week's issue.

PR #359, opened 2023-08-15
clean up alarms

tnsardesai:
# Clean up alarms for mesh services

## JIRA

https://clever.atlassian.net/browse/INFRANG-5475

## Overview

In Mesh there is no concept of alarm source of Target, ELB and Total as all metrics are coming from envoy sidecar. We are soon going to merge a change which will block CI for applications if source field is present for InternalErrorAlarms and mesh_config for prod set to mesh_only.

You have to review this PR to see that the new alarms make sense. It was hard to script all the different alarm setups we have across services. In most cases I have removed duplicate alarms but you might want to remove some more alarms or keep some deleted alarms.

## Rollout

Merge PR and deploy via dapple.

### DIFF
--- a/launch/workflow-manager.yml
+++ b/launch/workflow-manager.yml
@@ -30,30 +30,18 @@ databases:
 - dynamodb:us-west-1:workflow-manager-prod-v3
 alarms:
 - type: InternalErrorAlarm
-  severity: minor
-  parameters:
-    threshold: 0.01
-  extraParameters:
-    source: Target
-- type: InternalErrorAlarm
   severity: major
   parameters:
     threshold: 0.05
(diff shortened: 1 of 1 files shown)


launch/workflow-manager.yml
```diff
@@ -29,31 +29,19 @@ team: eng-infra
...
-    threshold: 0.01
-  extraParameters:
-    source: Target
 - type: InternalErrorAlarm
   severity: major
   parameters:
     threshold: 0.05
-  extraParameters:
-    source: Target
 - type: InternalErrorAlarm
```
taylor-sutton, line 36: This alarm should be minor

PR #364, opened 2023-08-23
FLARE-1520 - Conditionally fetch workflow history

andruwm:
## Link to JIRA:
[Link to JIRA](https://clever.atlassian.net/browse/FLARE-1520)

## Overview:
This PR adds conditional fetching on workflow history to the GetWorkflowByID endpoint. This history is not always needed, and fetching it is expensive. 

- [x] Run "make generate"


gen-go/server/db/tests/tests.go-e
```diff
@@ -3,14 +3,17 @@ package tests
...
 	"fmt"
-	"github.com/aws/aws-sdk-go/aws"
 	"testing"
 
 	"github.com/Clever/workflow-manager/gen-go/models"
 	"github.com/Clever/workflow-manager/gen-go/server/db"
 	"github.com/stretchr/testify/require"
 )
 
+func pointerToString(str string) *string {
```
taylor-sutton, line 13: Not your problem but a note: This whole file seems like it's not supposed to be here, something wrong on https://github.com/Clever/workflow-manager/blob/b1f03b3e3214fe83fe0fef34c89516feacd971b7/Makefile#L69 - and I think we fixed this wag bug anyway

taylor-sutton (approved):

gen-go/server/db/tests/tests.go-e
taylor-sutton: Actually I think this is why the build is failing?

PR #366, opened 2023-09-18
Increase timeout and batch size

andruwm:
## Overview:
These changes were made manually as part of a flare, and the lambda has since been frozen. This PR encodes the changes into the LC so the service can be deployed normally again.


ChrisScotMartin (approved):
👍

PR #376, opened 2024-01-19
[INFRANG-5393]: embedded workflow manager: race condition allows more…

philippeclever:
…-than-allowed concurrent executions

## Link to JIRA:
[Link to JIRA](https://clever.atlassian.net/browse/INFRANG-5393)

## Overview:
Fix issue by moving the goroutine inside the atomic count. This way, we block the race condition, and the routine works as expected

- [x] Run "make generate"

## Testing
manual to validate the race condition is not there anymore

### DIFF
--- a/embedded/poll.go
+++ b/embedded/poll.go
@@ -58,33 +58,32 @@ func (e *Embedded) pollGetActivityTask(ctx context.Context, resourceName string,
 			continue
 		}
 		// short circuit at the configured limit
-		if atomic.LoadInt64(concurrentExecutions) >= e.concurrencyLimits[resourceName] {
-			continue
-		}
-		select {
-		case <-ctx.Done():
-			log.Info("getactivitytask-stop")
-		default:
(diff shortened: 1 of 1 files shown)


taylor-sutton (approved):

ChrisScotMartin:
So I'm super curious about this one as I'm not seeing an actual logic change here. 

What I'm seeing is a change from:
```
for stuff
  if a >= b 
    continue
  do things
  call go routine
    go routine adds to concurrent executions counter
```
to
```
for stuff
  if a < b 
    do things
    call go routine
      go routine adds to concurrent executions counter
```

But I would think that in order to solve this race condition we would at least need to move the add to executions counter out of the go routine like:
```
for stuff
  if a >= b 
    continue
  do things
  add to concurrent executions counter
  call go routine
```

That being said I'd love to see this function get added to tests such that we can just run a `go test -race ./embedded`

PR #385, opened 2024-07-22
[INFRANG-6284] Fix unpropegated errors in sfn-execution-events-consumer

andruwm:
## Link to JIRA:
https://clever.atlassian.net/browse/INFRANG-6284

## Overview:
Fixes the bug causing errors not to be bubbled up

## Testing
Tested by running the lambda locally


tnsardesai (approved):
:shipit:

PR #386, opened 2024-07-29
Enhancements to workflow-ops scripts

taylor-sutton:
## Link to JIRA:
[Link to JIRA](https://clever.atlassian.net/browse/INFRANG-6265)

## Overview:
1. Allow setting the rate limit through command line flag and increase the default. I can get up to 100ish (more for short e.g. one-state workflows) on my laptop. But at that scale it will create DB throttling unless the DB is warmed up, so starting with 38 in the default.

2. Wait-ish for workers to be done before exiting.

If you made any changes to swagger.yml: N/A

## Testing

I ran this stuff for FLARE-1667


ChrisScotMartin (approved):
Nice, so already battle tested!

